Issue #121, opened 2015-01-11
Divide subparsers by propMetaVar in help


d12frosted:
Hey.

For example I have 

```haskell
progParser :: Parser Prog
progParser = 
  subparser $
  metavar "PROG" <>
  ...

cmdParser :: Parser Command
cmdParser = 
  subparser $
  metavar "CMD" <>
  ...

mainParser :: Parser (Prog, Command)
mainParser = (,) <$> progParser <*> cmdParser

opts :: ParserInfo Args
opts = info (helper <*> mainParser) idm
```

Which brings me to

```
> :set args --help
> execParser opts
Usage: <interactive> PROG CMD

Available options:
  -h,--help                Show this help text
  -c,--config FILE         Use configuration FILE instead of default one at
                           ~/.kitsunebook.cfg

Available commands:
  prog1                    Prog 1 description
  prog2                    Prog 2 description
  ..
  progn                    Prog n description
  cmd1                     Command 1 description
  cmd2                     Command 2 description
  ..
  cmdn                     Command n description
```

But I really would like to have the next help message:

```
Usage: <interactive> PROG CMD

Available options:
  -h,--help                Show this help text
  -c,--config FILE         Use configuration FILE instead of default one at
                           ~/.kitsunebook.cfg

Available PROGs (or progs):
  prog1                    Prog 1 description
  prog2                    Prog 2 description
  ..
  progn                    Prog n description

Available CMDs (or cmds):
  cmd1                     Command 1 description
  cmd2                     Command 2 description
  ..
  cmdn                     Command n description
```

I want this only for situation when I specify some custom `metavar`s.

Is it possible out of box?

joeyh:
FWIW, I have a similar wish for the list of subcommands. I have lots of subcommands, that naturally fall into different sets, and it'd be great if those sets could be displayed together.

nomeata:
+1

nomeata:
> FWIW, I have a similar wish for the list of subcommands. I have lots of subcommands, that naturally fall into different sets, and it'd be great if those sets could be displayed together.

Here is a hack for that, producing this output:

```
Available commands:
  GME creation commands:   
  assemble                 creates a gme file from the given source
                           
  OID code creation commands:
                           
  oid-table                creates a PDF file with all codes in the yaml file
  oid-codes                creates PNG files for every OID in the yaml file.
  oid-code                 creates PNG files for each given code(s)
                           
  GME analysis commands:   
[...]

PR #122, opened 2015-01-11
Divide subparsers by propMetaVar in parserHelp

d12frosted:
Fixing #121

Tried to make clean fix, but run into problems with gather function (it is needed to cover situations when library user defines few subparsers without providing metavar or defines few subparsers with same metavar). 

OK, what we have after this update. Suppose I have written:

```haskell
cmd :: String -> Parser a -> String -> Mod CommandFields a
cmd s p d =
  command s (info p (progDesc d))

parser1 :: Parser Int
parser1 =
  subparser $
  metavar "INT" <>
  cmd "one" (pure 1) "Got 1" <>
  cmd "two" (pure 2) "Got 2"

parser2 :: Parser Char
parser2 =
  subparser $
  metavar "CHAR" <>
  cmd "lol" (pure 'a') "Got a" <>
  cmd "hel" (pure 'b') "Got b"

parser3 :: Parser Bool
parser3 =
  subparser $
  cmd "no" (pure True) "Got true" <>
  cmd "yes" (pure False) "Got false"

parser4 :: Parser (Maybe Bool)
parser4 =
  subparser $
  cmd "just yes" (pure (Just True)) "Got Just true" <>
  cmd "nothing" (pure Nothing) "Got nothing"

test :: ParserInfo (Int, Char, Bool, Maybe Bool)
test = info (helper <*> ((,,,) <$> parser1 <*> parser2 <*> parser3 <*> parser4)) idm
```

And when I do

```
> :set args --help
> execParser test
```

I got fancy result:

```
Usage: <interactive> INT CHAR COMMAND COMMAND

Available options:

Available commands:
  no                       Got true
  yes                      Got false
  just yes                 Got Just true
  nothing                  Got nothing

Available chars:
  lol                      Got a
  hel                      Got b

Available ints:
  one                      Got 1
  two                      Got 2
*** Exception: ExitSuccess
```

The only problems I actually see are:

* using `metavar "OPTION"` is allowed (so user will see "Available options:" twice) - it's not that critical and can be easily avoided
* using `metavar "CHARS"` will lead to "Available charss:". I see few solutions here:
  * not adding 's' character to the end of title, when it already has one there
  * not lowering title, so in this case user will see "Available CHARSs:"
  * not adding 's' character to custom titles at all [+ not lowering title]

I leave this questions open (so maintainers can use best solution they think of). 

BTW, after running tests I've got

```
All 43 tests passed (0.67s)
```

### DIFF
--- a/Options/Applicative.hs
+++ b/Options/Applicative.hs
@@ -34,5 +34,3 @@ import Options.Applicative.Common
 import Options.Applicative.Builder
 import Options.Applicative.Builder.Completer
 import Options.Applicative.Extra
-
-{-# ANN module "HLint: ignore Use import/export shortcut" #-}
--- a/Options/Applicative/Builder.hs
+++ b/Options/Applicative/Builder.hs
@@ -44,6 +44,7 @@ module Options.Applicative.Builder (
   hidden,
   internal,
   command,
+  command',
   completeWith,
   action,
   completer,
@@ -107,6 +108,7 @@ import Options.Applicative.Common
 import Options.Applicative.Types
 import Options.Applicative.Help.Pretty
 import Options.Applicative.Help.Chunk
+import Data.List (intercalate)
 
 -- readers --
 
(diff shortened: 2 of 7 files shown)


d12frosted:
Oh, and one more thing.
Probably you want to change `gather` function to

```haskell
gather :: [(String, Chunk Doc)] -> [(String, Chunk Doc)]
gather = reverse . flip gather' []
```

So order of cmdDesc is natural.

HuwCampbell:
I've been thinking about this. This PR is actually tacking a few things, one or two of which I'm not too keen on.
I am not convinced that hidden commands should exist, because they are used on the top level as the main parser, and hiding that parser makes no sense to me. One can however hide subparsers (using internal or hidden) which can be used for aliases and the like.
I agree with your concerns around plural forms, it's going to be incorrect for a lot of people.

But, you got me thinking, in the original, we have
```
cmdDesc = vcatChunks . mapParser desc
```
If this were changed to 
```
cmdDesc = vsepChunks . mapParser desc
```
Then the commands from different subparsers would have a line separation between them, which would more or less do what you're asking for — separate the commands based on subparser (or metavar, although it wouldn't regather if two metavars were the same).

HuwCampbell:
Please see #197 for an implementation allowing this, I think it's a bit more flexible and will be more backwards compatible.

HuwCampbell:
I'm going to close this one as #197 covers it off. Thanks though, it helped me design said change.

d12frosted:
Thanks. I had no chance to check your proposed solution, because I don't use this lib anymore. But hopefully one day I'll give it a try 💃 💯

PR #123, opened 2015-01-13
Allow for unix style short args, i.e., -xcf

HuwCampbell:
Most command parsers (getOpts and the like) allow for unix style short commands, which can be concatenated together, for example, tar -xcf.

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -119,8 +119,11 @@ optMatches disambiguate opt (OptWord arg1 val) = case opt of
       lift $ runReadM (withReadM (errorFor arg1) (crReader rdr)) arg'
   FlagReader names x -> do
     guard $ has_name arg1 names
-    guard $ isNothing val
-    Just $ return x
+    Just $ do
+      args <- get
+      let val' = (\s -> '-' : s) <$> val
+      put $ maybeToList val' ++ args
+      return x
   _ -> Nothing
   where
     errorFor name msg = "option " ++ showOption name ++ ": " ++ msg
--- a/tests/Tests.hs
+++ b/tests/Tests.hs
@@ -351,6 +351,37 @@ case_arg_order_3 = do
     Success res -> (3, 5) @=? res
     _ -> assertFailure "unexpected parse error"
 
+case_unix_style :: Assertion
+case_unix_style = do
+  let p = (,)
+          <$> flag' (1 :: Int) (short 'x')
+          <*> flag' (2 :: Int) (short 'c')
+      i = info p idm
+      result = run i ["-xc"]
+  case result of
+    Success res -> (1, 2) @=? res
+    _ -> assertFailure "unexpected parse error"
+
+case_unix_with_options :: Assertion
+case_unix_with_options = do
+  let p = (,)
+          <$> flag' (1 :: Int) (short 'x')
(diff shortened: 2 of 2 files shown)


FrontierOtr:
Bump.

fosskers:
+1 for this.

pcapriotti:
I had an objection about this, but I forgot what it was. I'll probably incorporate it in the next release.

fosskers:
It would be much appreciated.

mrak:
The only objection I can think of is ambiguous parsing between a shortflag adjacent to it's argument versus declaring multiple parameter-less flags together.

```sh
program -sarg
# same as
program -s arg
```

versus

```sh
program -sarg
# same as
program -s -a -r -g
```

I would argue that the second example is more often the intended behavior. Some clever parsing may be needed to support backward compatibility with the first example

pcapriotti:
I haven't looked at the patch yet, so I'm not sure what this does in this case, but it should be possible to support both behaviors: the first parse should be used for regular options, and the second for flags. This should resolve the ambiguity. But maybe I'm missing something.

fosskers:
Is current behaviour not:
```
program -s=arg
```
Certainly that would remove ambiguity.

pcapriotti:
No, that syntax is now allowed. You can write either:

    program -s arg

or

    program -sarg

for short names.

See https://github.com/pcapriotti/optparse-applicative#regular-options for details.

FrontierOtr:
I wrote is so that -sarg is the same as -s -a -r -g.
The first case would be a little more complicated as the flags are put into OptWords before one knows whether they are options or flags.
Currently the '=' syntax only works for long options, e.g., --file=name.psv

pcapriotti:
> I wrote is so that -sarg is the same as -s -a -r -g.

I don't want to change the behaviour of existing code. If a regular option has the short name 's', then `-sarg` should be parsed as `-s arg`.

FrontierOtr:
Hmm, yeah fair on the backwards compatibility. 
I posit that the way to do it then is, inside the FlagParser, instead of 
    guard $ isNothing val
placing val if it exists, prepended be a single '-', inside the StateT.

FrontierOtr:
Something like this:
https://github.com/markhibberd/pirate/pull/34

FrontierOtr:
Ok, I have implemented it in a way which shouldn't break compatibility.
Options will consume the argument, while flags will take the head and create a new set of short options to be parsed.

FrontierOtr:
@pcapriotti I think this current solution is pretty neat, and it doesn't break anything. Please do have a look.

FrontierOtr:
Sorry if I'm being a pain in the arse. I think this solution is reasonable, could you please comment.

pcapriotti:
Sorry, I have been busy with other things lately. I'll take a look as soon as I have some free time.

voidzero:
+1 for this. Or +2 if I may.

jystic:
:+1:

HuwCampbell:
Just another prod since it's been a month. Also I have a PR which improves the help text markedly in #139

LightAndLight:
I was just trying to figure out how to hack this functionality. Keen

fosskers:
This is excellent news.

jystic:
:tada:

PR #125, opened 2015-02-06
Remove PaserM

larskuhtz:
I don't expect this PR to be merged upstream. The main motivation for PR is to understand the reason for the existence of `PaserM` and `BindP`.

I understand that there may be situations where it is convenient to have monadic bind when writing parsers. However, having used optparse-applicative a lot in many projects, I never encountered a situation where I actually ended up using `PaserM`.

I find that the current code in `Options.Applicative.Types` is difficult to understand due to the circular (and as it seems to me redundant) dependency between `PaserM` and the `Alternative` instance of `Parser`. Also it feels like `MultP` shouldn't be needed in the presence of `BindP`.

This PR completely removes `ParserM` and `fromM`, `oneM`, `manyM`, `someM`. So it's an API change that may break existing code. But even with `ParserM`, I wonder why it is used in the implementation of `Alternative` for `Parser`, and why `BindP` and `MultP` are kept both.

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -161,10 +161,6 @@ searchParser f (MultP p1 p2) = foldr1 (<!>)
 searchParser f (AltP p1 p2) = msum
   [ searchParser f p1
   , searchParser f p2 ]
-searchParser f (BindP p k) = do
-  p' <- searchParser f p
-  x <- hoistMaybe (evalParser p')
-  return (k x)
 
 searchOpt :: MonadP m => ParserPrefs -> OptWord -> Parser a
           -> NondetT (StateT Args m) (Parser a)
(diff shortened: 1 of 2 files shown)


larskuhtz:
The changes in this PR seem to cause the parser to diverge. I guess this is due to removing the implementations of `many` and `some` from the `Alternative` instance from `Parser`. If that's actually the case, shouldn't explicit implementations have the same semantics as the default implementations?

pcapriotti:
Yeah, the `BindP` stuff is there to make `many` and `some` work. I know that the documentation says that every implementation must be equal to the default one, but that cannot possibly be achieved in this case. But I don't want to define separate functions `many'` and `some'`, so I think this is a necessary evil.

See also the discussion [here](http://www.reddit.com/r/haskell/comments/2j8bvl/laws_of_some_and_many/).

PR #126, opened 2015-02-07
Fix link to blog post

begriffs:
The github pages settings for paolocapriotti.com require it to be prefixed by `www`. So you could merge this PR or modify your site [CNAME](https://github.com/pcapriotti/pcapriotti.github.com/blob/master/CNAME) to remove the `www`. Not sure of all the side effects of the latter but I have my own github pages site configured without the `www` and it seems to work fine.


pcapriotti:
I updated the IP in the DNS record. It seems to work fine now. Thanks.

PR #134, opened 2015-04-06
Add defaultPrefs, a default preferences value

bos:
I was mystified as to how to use `execParserPure` and related functions
to debug a parser, because I couldn't see where there existed a way
to specify default preferences.

Turns out such a way didn't actually exist! This is much preferable
to using `prefs idm` directly, as the word "default" is the first
thing a user will search for, whereas I stumbled on `prefs idm` only
by reading the source.


pcapriotti:
Thanks!

Issue #112, opened 2014-11-04
Infinite loop when generating parsers from free alternative functor


nc6:
Appreciate this is probably an unusual use case - I have a schema which is used to generate multiple parsers for configuration options, implemented as a free alternative over an `Option` functor (using the `free` package). I would like to generate an optparse-applicative parser from this, but it leads to an infinite loop when converting `many` and `some` expressions, I believe because optparse overloads `many` and `some` such that they're embodying more than just alternative parsing. 

Is this a bug or a deliberate feature? I can't see a way to get around this, which would be a shame as it seemed one reason why optparse-applicative was such a good match.

Minimal code example at http://lpaste.net/113737, relevant stack overflow question at http://stackoverflow.com/questions/26718443/generate-optparse-applicative-parser-from-free-alternative-functor

pcapriotti:
Closing, as there is nothing actionable from this.

PR #136, opened 2015-05-29
Implement 'hoist' in various places.

nc6:
Not sure whether you're still working on this branch, but I finally got the time to start doing as you suggested in https://github.com/pcapriotti/optparse-applicative/issues/112, and needed to implement `hoist`, so thought I'd push the changes upstream.


HuwCampbell:
I don't think this is going anywhere right now unfortunately. Optparse is essentially stable right now, so even though the branch adoption may happen in the future, I don't think its lineage will be direct enough to incorporate this PR.

Thanks for contributing, we do appreciate it.

PR #139, opened 2015-06-14
Topic/describe missing

HuwCampbell:
This changes the error text to have, for example:
Missing: -a ARG
when an item when -a is not supplied but is required. It works across the whole tree so if one of two alternatives are needed but not supplied, then the error text will be something like
Missing: (-a | -b)


nhibberd:
:+1:

jystic:
:+1:

PR #147, opened 2015-08-11
Add Text version of strOption and strArgument

markus1189:
Adds:

- `txt` a `ReadM` for text, analog to `str`
- `txtOption` and `txtArgument` analog to `strOption` and `strArgument` for String

I had to **add `text` as a dependency**, I am not sure if we need a version bound at all, but I picked the one from the `commercialhaskell/stack` project (arbitrarily).

And for bonus points, here is the effect of adding `text` on the dependency graph, showing that it only adds `text`:
optparse

optparse-with-text

### DIFF
--- a/Options/Applicative/Builder.hs
+++ b/Options/Applicative/Builder.hs
@@ -19,6 +19,7 @@ module Options.Applicative.Builder (
   -- creates a parser for an option called \"output\".
   subparser,
   strArgument,
+  txtArgument,
   argument,
   flag,
   flag',
@@ -26,6 +27,7 @@ module Options.Applicative.Builder (
   abortOption,
   infoOption,
   strOption,
+  txtOption,
   option,
   nullOption,
 
@@ -58,6 +60,7 @@ module Options.Applicative.Builder (
   -- | A collection of basic 'Option' readers.
   auto,
   str,
+  txt,
   disabled,
   readerAbort,
   readerError,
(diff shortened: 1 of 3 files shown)


HuwCampbell:
Having a Text ReadM sounds quite reasonable to me; locking to a specific version though would probably break a lot of people's applications.
I'll see if I can figure something out for the major next release.

markus1189:
FYI: rebased the pr onto the current master :)
Thanks for maintaining optparse-applicative.

phadej:
-  http://hackage.haskell.org/package/text-0.8.0.0 uploaded *Tue Aug 31 22:37:25 UTC 2010* has `Text` and `pack`. So the lower bound can be quite low.
- http://matrix.hackage.haskell.org/package/text shows that `>=1.2.0.4` is totally reasonable.
- lts-2.22 (the last for GHC7.8) has https://www.stackage.org/lts-2.22 (someone might still have that as their base snapshot, e.g. me)

Btw. rebase needed again.

markus1189:
@phadej so I guess that means `>= 1.2.0.4` is okay?

Rebased onto latest master :+1:

HuwCampbell:
I had a think about this, I believe that we can get the benefit without the dependency by making `str` polymorphic over the `IsString` class. Everything else kind of falls out for free.
See #223

markus1189:
:+1: seems like a good idea

PR #150, opened 2015-08-26
Relax dependency on the process package

sapek:
Version 1.3 of process package has been released and the dependency
process < 1.3 causes conflicts in some cases and can be relaxed.

I verified that the tests build and pass with `process` 1.3.

### DIFF
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -1,5 +1,5 @@
 name:                optparse-applicative
-version:             0.11.0.2
+version:             0.11.0.3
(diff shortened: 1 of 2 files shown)


HuwCampbell:
Seems reasonable. I might do the version bumping myself though if that's ok.

sapek:
Sounds good, `tasty` has only (>=) dependency.

HuwCampbell:
Cherry-picked in. Thanks!

sapek:
Thanks. Could you please add a comment to the PR when you publish to hackage? I want to remove my workaround from Bond then.

PR #162, opened 2015-10-14
Replace WriterT in P Monad stack with StateT.

HuwCampbell:
This allows us to enter and exit the context more easily, and reset
it when a parser is finished.

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -52,7 +52,7 @@ module Options.Applicative.Common (
   OptDescStyle (..)
   ) where
 
-import Control.Applicative (pure, (<*>), (<*), (<$>), (<|>), (<$))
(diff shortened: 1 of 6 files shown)


Options/Applicative/Common.hs
```diff
@@ -102,14 +102,14 @@ argMatches opt arg = case opt of
     return result
   CmdReader _ f ->
     flip fmap (f arg) $ \subp -> StateT $ \args -> do
-      setContext (Just arg) subp
+      enterContext arg subp
```
phadej: Should there be `bracket` like abstraction for entering/exiting the context

HuwCampbell: Wouldn't a bracket exit the context on failure as well though? The point is that on failure we still have it.

HuwCampbell: I've changed it to be more clear that it's wrapping this command.

PR #180, opened 2016-01-13
Feature/enum

aloiscochard:
Hi there, 

I was looking for automatic support for simple ADT (aka Enum), more specially, dataconstructors have to start with a capital letter but in the unix world we usually pass values lowercase.

So this `enum` just does that, you can see it in action here:
http://pastie.org/10687143

We could later improve or add more related functionalities (like outputing the alternatives in the help), but we can start with this simple combinator.

I understand I might have missed something that would allow me to achieve what I want, but I could not find an other way than writing this combinator.

Thanks for this great library.

### DIFF
--- a/Options/Applicative/Builder.hs
+++ b/Options/Applicative/Builder.hs
@@ -58,6 +58,7 @@ module Options.Applicative.Builder (
   -- | A collection of basic 'Option' readers.
   auto,
   str,
+  enum,
   disabled,
   readerAbort,
   readerError,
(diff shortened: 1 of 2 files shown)


HuwCampbell:
I think you'd want `Bounded` instead of `Enum` – `[minBound .. maxBound]` would be significantly safer, as `toEnum` throws exceptions (and there's no real guarantee there's a 0, or that 0 is the lowest element), and it would be too easy to cause a hang or very long parse times as enumerations can be infinite.

But I'd still a bit hesitant, this could be very slow. One should probably be using a parser and `eitherReader`; or for simple sum types, one can derive `Read` and use `auto`.

HuwCampbell:
Thanks for submitting this PR, but, I don't think this is safe enough to accept into the main library. There are too many edge cases, and seeing enum or bounded might mean people try to use it with Int or any other large datatype.
As I said, eitherReader is your friend here.

PR #181, opened 2016-01-15
Allow building with transformers 0.5 (on GHC 8 RC)

Heather:


23Skidoo:
+1. Would be also nice if the upper bound was bumped for the released version on Hackage.

HuwCampbell:
Hi, thanks.
Not sure if this is correct however, as there's a requirement for transformers-compat in the 0.4 branch that transformers is < 0.5. So this probably won't resolve the way you're expecting.
Ed is preparing for a new version over there (https://github.com/ekmett/transformers-compat/commit/4a29c5cd3ec1c67fdeb0bbd0172ede046b5ac18c).
Once that gets pushed through I'll bump it all up.

Heather:
@HuwCampbell to be honest I don't see actual problem with transformers-compat, I used to install it and used it to build another application...

HuwCampbell:
The problem is that optparse depends on transformers-compat, which currently requires transformers to be < 0.5 (although this will change very soon with Ed's changes).
So one still can't use transformers 0.5 after this change. To demonstrate, I can set `transformers == 0.5.*` – then I get a conflict:
```
trying: optparse-applicative-0.12.0.0 (user goal)
next goal: transformers (dependency of optparse-applicative-0.12.0.0)
rejecting: transformers-0.3.0.0/installed-645... (conflict:
optparse-applicative => transformers==0.5.*)
trying: transformers-0.5.0.0
trying: transformers-compat-0.4.0.4 (dependency of
optparse-applicative-0.12.0.0)
trying: transformers-compat-0.4.0.4:-two
rejecting: transformers-compat-0.4.0.4:-three (conflict:
transformers==0.5.0.0, transformers-compat-0.4.0.4:three =>
transformers>=0.4.1 && <0.5)
rejecting: transformers-compat-0.4.0.4:+three (manual flag can only be changed
explicitly)
Dependency tree exhaustively searched.
```

Heather:
@HuwCampbell oh right it's just bumped so `transformers-compat` should be `< 0.6` either, I was making this PR before version bump

HuwCampbell:
Merged a similar PR, thanks!

PR #189, opened 2016-02-15
Allow more complex usage of many and some.

HuwCampbell:
Previously the entire BindP was required to pass during each call to
searchParser. Meaning that many pairs of arguments were not possible. By
being a bit more lazy in progressing through these options, and rebuilding
the BindP instead of asserting that it must be complete, we can allow for
many pairs or n-tuples (or other complex constructed datatypes) of items.

For example:
```haskell
p = many $ (,) <$> argument str idm <*> argument str idm
```

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -168,9 +168,7 @@ searchParser f (BindP p k) = msum
        return $ BindP p' k
   , case evalParser p of
       Nothing -> mzero
-      Just aa -> do
-        k' <- searchParser f (k aa)
-        return k' ]
+      Just aa -> searchParser f (k aa) ]
 
 searchOpt :: MonadP m => ParserPrefs -> OptWord -> Parser a
           -> NondetT (StateT Args m) (Parser a)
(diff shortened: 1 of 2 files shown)


ibotty:
One thing that could be enhanced is the usage help.

For `many $ (,) <$> option str (short 'a') <*> option str (short 'b')` it will yield

    Usage: <program> [-a ARG] [-b ARG]

I would expect

    Usage <program> [-a ARG [-b ARG]]...

because the the options are dependent. (The `...` part is of course a different feature request.)

HuwCampbell:
Thanks, glad I could help. More tests are always welcome.

Yep, I'm aware of the usage text issues. It will require a bit more work still, as the current `OptTree` type is not designed to handle cases like these.

HuwCampbell:
Closing as this PR is not over master

ibotty:
Sorry for being a little dense. Is it merged? I tested this patch in a program of mine and it works wonderfully.

ibotty:
The answer is yes. Thank you for the prompt patch!

PR #149, opened 2015-08-23
Fix GHC 7.10 warnings.

23Skidoo:

PR #192, opened 2016-03-10
Fix various 7.10 warnings around Monoid and Applicative.

tmcgilchrist:
Updated version of this fix using the `import Prelude` solution

https://github.com/pcapriotti/optparse-applicative/pull/149


HuwCampbell:
Looks good to me

PR #197, opened 2016-04-17
Basic grouping of subparser commands

HuwCampbell:
This exposes a new modifier such that groups of commands can have their own textual description, such as Dangerous commands:, or "Infrequent commands:".


mgsloan:
I haven't reviewed this carefully, but +1 on the overall idea

PR #213, opened 2016-07-17
Split up Monoid instances into Semigroup and Monoid

bgamari:


HuwCampbell:
G'day,

I think that doing this is a good thing, but I'm a bit hesitant to push across all GHC versions, especially as the semigroups monoid situation is in a bit of flux.
You've probably got thoughts about this, based on your reviewing of the changes for ghc.

Can you see any way that this could create problems for users on 7.8 or 7.10? Personally, I would prefer to play it safe, and also not incur the semigroups dependency.

I was thinking a good middle ground might be to CPP in the semigroup instance for GHC 8 and higher?

Thanks,
Huw

bgamari:
Huw Campbell <notifications@github.com> writes:

> G'day,
>
Hi!

> I think that doing this is a good thing, but I'm a bit hesitant to
> push across all GHC versions, especially as the semigroups monoid
> situation is in a bit of flux.
> 
> You've probably got thoughts about this, based on your reviewing of
> the changes for ghc.
>
> Can you see any way that this could create problems for users on 7.8
> or 7.10? Personally, I would prefer to play it safe, and also not
> incur the semigroups dependency.

The proposal was specifically designed with compatibility in mind, so no
I can't see any reason why adding these instances would pose a problem
for older compilers. The Core Libraries Committee's official
recommendation on this matter can be found here [1]. The official stance
is to add a dependency on `semigroups`.

> I was thinking a good middle ground might be to CPP in the semigroup
> instance for GHC 8 and higher?

The problem with this approach is that users who want to use `Semigroup`
with older compilers can no longer assume that the necessary instances
will be available in `optparse-applicative`. This would likely be quite
annoying for users as it would mean that they would potentially need
both `Data.Monoid.(<>)` and `Data.Semigroup.(<>)` in scope to use
`optparse-applicative`.

Cheers,

- Ben


[1] https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid

Issue #211, opened 2016-06-19
Behavior of double dash ('--') is surprising when there are no positional argument parsers


ExpHP:
If a parser contains only `switches` and `strOptions` but lacks `strArguments` (or all of its `strArguments` have been parsed), then its behavior with '--' appears to be the following:

 * The first `--` is gobbled.
 * **Arguments after the first `--` continue to be parsed by the switches and strOptions**.
 * A second `--` generates an error.

I find the second point to be surprising because I would expect these arguments to be regarded as "positionals" regardless of whether there are any `strArguments` waiting to consume them.

**Demonstration:**

    import Control.Applicative
    import Options.Applicative
    
    -- Usages:
    --   main1 [-a]
    --   main2 [-a] Arg
    --   main3 [-a] [Arg...]
    data Args1 = Args1 Bool deriving Show
    data Args2 = Args2 Bool String deriving Show
    data Args3 = Args3 Bool [String] deriving Show
    
    parser1 = Args1 <$> switch (short 'a')
    parser2 = Args2 <$> switch (short 'a') <*> strArgument (metavar "Arg")
    parser3 = Args3 <$> switch (short 'a') <*> (many.strArgument) (metavar "Arg...")
    
    -- compile as main1, then subsitute 'parser1'
    --    here to compile main2 and main3
    theParser = parser1
    main = execParser (info theParser idm) >>= print

**Comparison of actual output and what I expected to see:**

    Commands:
        ./main1 -- -a
        ./main2 -- -a
        ./main3 -- -a
    Output:
        Args1 True
        Args2 False "-a"
        Args3 False ["-a"]
    Expected to see:
        Invalid argument `-a'      (mismatch)
        Args2 False "-a"           (OK)
        Args3 False ["-a"]         (OK)

    Commands:
        ./main1 -- -- -a
        ./main2 -- -- -a
        ./main3 -- -- -a
    Output:
        Invalid option `--'
        Args2 True "--"
        Args3 False ["--","-a"]
    Expected to see:
        Invalid argument `--'      (minor mismatch)
        Invalid argument `-a'      (mismatch)
        Args3 False ["--","-a"]    (OK)

HuwCampbell:
You are correct. The behavior after `--` is that arguments are searched first, then, if none are found, options and flags are [considered](https://github.com/pcapriotti/optparse-applicative/blob/master/Options/Applicative/Common.hs#L197-L199).
[...]

PR #214, opened 2016-07-20
Don't allow options after '--' if there are no arguments

HuwCampbell:
Related to #211
Treating arguments after a '--' if there's no arguments left to parse as options could lead to anomalous behaviour in some cases.


HuwCampbell:
The names for the constructors `SkipOpts` and `AllowOpts` don't make a lot of sense with this change unfortunately.

PR #216, opened 2016-08-01
Semigroup

HuwCampbell:
On top of #213

### DIFF
--- a/Options/Applicative.hs
+++ b/Options/Applicative.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE CPP #-}
 module Options.Applicative (
   -- * Applicative option parsers
   --
--- a/Options/Applicative.hs
+++ b/Options/Applicative.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 module Options.Applicative (
   -- * Applicative option parsers
   --
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -93,6 +93,11 @@ source-repository head
   location: https://github.com/pcapriotti/optparse-applicative.git
 
 library
+  if impl(ghc >= 8)
+    ghc-options:       -Wall -Wno-redundant-constraints
+  else
+    ghc-options:       -Wall
+
   exposed-modules:     Options.Applicative,
                        Options.Applicative.Arrows,
                        Options.Applicative.BashCompletion,
@@ -109,14 +114,15 @@ library
                        Options.Applicative.Types,
                        Options.Applicative.Internal
 
-  ghc-options:         -Wall
 
   build-depends:       base                            == 4.*
-                     , semigroups                      >= 0.10 && < 0.19
                      , transformers                    >= 0.2 && < 0.6
                      , transformers-compat             >= 0.3 && < 0.6
                      , process                         >= 1.0 && < 1.5
                      , ansi-wl-pprint                  >= 0.6.6 && < 0.7
+
+  if !impl(ghc >= 8)
+    build-depends:     semigroups                      >= 0.10 && < 0.19
 
 
 test-suite optparse-applicative-tests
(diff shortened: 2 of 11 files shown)


HuwCampbell:
@bgamari I have built up a new PR on top of your work, hope you don't mind.

The original PR probably may have broken a few things for people, given that we re-export `Data.Monoid`'s `(<>)` operator, and in your version it was changed to exporting `Data.Semigroup`'s `(<>)`. In 8.2 these will hopefully be identical, but for now it could cause a conflict if people are importing `Data.Monoid` for something else.

This is one of the reasons the test suite was failing over in #213. Please have a look, let me know if you see any issues.

bgamari:
@HuwCampbell sounds like a sensible way forward.

@hvr, do you have an opinion on this?

Options/Applicative.hs
```diff
@@ -14,6 +15,11 @@ module Options.Applicative (
   -- | The standard @Applicative@ module is re-exported here for convenience.
   module Control.Applicative,
 
+  -- | Re-export Data.Monoid's (<>) operator for convenience.
```
hvr: This is a bit evil as then optparse-applicative's exposed API depends on its transitive dependencies configuration. I.e., if you happen to use `base-4.5` or later (even if you don't declare direct dependency on `base`) you get `(<>)` in scope, if use anything else, importing `Options.Applicative` does not provide `((<>))`. 

So this is nasty API-variability which causes us Hackage Trustees quite some headaches, as we'd need to add constraints that are way beyond the kind of things we can edit into `.cabal` files retroactively.

I'd strongly advise against this. If you want to export conditionally, better drop support for base < 4.5 and thereby avoid the conditional  export altogether :-)

(or you could provide a local `<>` definition for when `base` doesn't provide it; this runs the risk of potential clashes with other local `<>` definitions but it's the lesser evil IMO)

optparse-applicative.cabal
```diff
@@ -112,6 +112,7 @@ library
   ghc-options:         -Wall
 
   build-depends:       base                            == 4.*
+                     , semigroups                      >= 0.10 && < 0.19
```
hvr: fwiw, you can guard this via `if !impl(ghc >= 8)` if you want to avoid redundant compat dependency. Otoh,  seems pointless, since there's a `transformers-compat` dep below, so that's probably not a huge concern here.

Options/Applicative.hs
bgamari: This is a bit evil as then `optparse-applicative`'s exposed API depends on its transitive dependencies configuration.

Oh dear, I misread @HuwCampbell's original message. Yes, I agree, this is not good. Users shouldn't have to worry about the exposed interface for a given version of a package changing beneath their feet. I rescind my claim that this is a sensible way forward.

@HuwCampbell, I thought your suggestion here was to merely continue to unconditionally expose `Data.Monoid`. Really, I'm still not convinced that we shouldn't simply add a dependency on `semigroups` for GHC <8.9 and export its `(<>)`.

HuwCampbell: Hmm. This was entirely to be be consistent with what `optparse-applicative` has been exporting since [0.7](https://hackage.haskell.org/package/optparse-applicative-0.7.0/docs/src/Options-Applicative-Builder.html).

That said, thank you both for the input. I'll probably drop the export completely and let the user decide which instance to use.

optparse-applicative.cabal
HuwCampbell: Ta

Options/Applicative.hs
bgamari: > That said, thank you both for the input. I'll probably drop the export completely and let the user decide which instance to use.

As a frequent user of the library I think that is a pretty good way forward. I often end up hiding `(<>)` from `Options.Applicative` anyways.

bgamari:
Thanks @HuwCampbell!

PR #223, opened 2016-08-19
Rewrite str, eitherReader and maybeReader to be polymorphic over IsSt…

HuwCampbell:
…ring.

This has significant benefits, in that it covers off Text and ByteString
readers for free (and without dependencies) and works pretty seamlessly
with Attoparsec parsers.

May cause breakages if people are using str for type inference.
In that case, adding type signatures should fix the problem.

### DIFF
--- a/tests/test.hs
+++ b/tests/test.hs
@@ -12,8 +12,10 @@ import qualified Examples.Formatting as Formatting
 
 import           Control.Applicative
 import           Control.Monad
+import           Data.ByteString (ByteString)
 import           Data.List hiding (group)
 import           Data.Semigroup hiding (option)
+import           Data.String
 
 import           System.Exit
 import           Test.QuickCheck hiding (Success, Failure)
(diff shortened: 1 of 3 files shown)


tests/test.hs
```diff
@@ -129,7 +131,8 @@ prop_alt_cont = once $
 
 prop_alt_help :: Property
 prop_alt_help = once $
-  let p = p1 <|> p2 <|> p3
+  let p :: Parser (Maybe (Either String String))
```
phadej: FWIW: this is breaking change, as the need to alter tests highlights, so would need a major version bump.

HuwCampbell: I'm aware, hence the comment
```
 -- Polymorphic on IsString since 0.14.
```

HuwCampbell: I will probably not release this change immediately anyway, but wait until I can roll up a few more fixes into a release.

PR #226, opened 2016-09-03
optparse-applicative.cabal: allow QuickCheck-2.10, add missing files to tarball

trofi:
Signed-off-by: Sergei Trofimovich <siarheit@google.com>


HuwCampbell:
Looks like an oversight on my part with those missing test cases. Hope it didn't cause you any issues.
Looks good. thanks!

PR #230, opened 2016-10-14
[Please Review] Expanding first example

JustusAdam:
I recently used this library for the first time, and I found myself using the first example as sort of a TL;DR rather than reading the whole readme. 

Now that I found out how to use values other than switches and Strings I thought it might be helpful to add such a common use case to the initial example, so that it is easier to find how to use arbitrary values.

Additionally the example now contains default values and short flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,8 @@ Here is a simple example of an applicative option parser:
 import Options.Applicative
 
 data Sample = Sample
-  { hello :: String
-  , quiet :: Bool 
+  { hello  :: String
+  , quiet  :: Bool 
   , repeat :: Int }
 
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Thanks! Seems like a good addition, one isn't actually required to write "(defaults to 1)" though, the library can do it itself with `showDefault`.

Maybe use that, as I don't think it's really mentioned anywhere else please.

JustusAdam:
Oh interesting. I expected something like that to exist, but I didn't find it. (Not that I looked particularly hard)

HuwCampbell:
Cool, thanks!

PR #232, opened 2016-10-16
fix missing imports in README.md

yuhangwang:
fix missing imports in the example from the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,8 @@ Here is a simple example of an applicative option parser:
 
 ```haskell
 import Options.Applicative
+import Data.Monoid -- will import <> (a.k.a. mappend)
+import Control.Applicative -- will import <*>, <$>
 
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Hey, thanks for this.
A few comments,  `Data.Semigroup` would be better, as it won't break when 8.2 comes along, and qualified import don't require documentation.
Cheers,
Huw

yuhangwang:
Thanks for the tip!

HuwCampbell:
Heh,
The M doesn't refer to the thing being repeated, but that it's repeating an "Monadic" action. It should actually be `replicateM_` from `Control.Monad` though. That section of the README was added in separate PR a few days ago and I didn't notice that the function name was incorrect.
Huw

HuwCampbell:
Have done the fix ups and rebased your PR. Thanks!

Issue #233, opened 2016-10-22
noIntersperse never parses options


Thhethssmuz:
I am trying to create a parser for a program that is a runner for another program, and I also want to pass arguments along to this other program. Ideally I want to do this without having to use the `--` option after the main `file` argument since I know that all options after this argument should just be passed along.

From what I can gather, from #22, this seams to be the intended use case for the `noIntersperse` mode. And from them Haddocks:

> `noIntersperse :: InfoMod a`
> 
> Disable parsing of regular options after arguments

I take this to mean that regular options should be parsed prior to any argument. However, it seams that this option does not parse options at all:

``` haskell
data Options = Options
  { foo  :: Bool
  , file :: String
  , args :: [String]
  } deriving Show

parser :: Parser Options
parser = Options
  <$> switch ( long "foo" )
  <*> strArgument ( metavar "FILE" )
  <*> many ( strArgument ( metavar "ARGS..." ) )

main :: IO ()
main = do
  opts <- execParser $ info (helper <*> parser) noIntersperse
  print opts
```

When I run the above code with `--foo myfile -a -b -c` I expect to get:

```
Options {foo = True, file = "myfile", args = ["-a","-b","-c"]}
```

but instead I get:

```
Options {foo = False, file = "--foo", args = ["myfile","-a","-b","-c"]}
```

It also breaks the `help` option:

```
Options {foo = False, file = "--help", args = []}
```

Am I missing something here?

HuwCampbell:
No you're not, I think this counts as a bug.

Currently turning on noIntersperse has the same behaviour as going after `--`, I'll have a think about a solution for you, and will add your test to the test suite.

HuwCampbell:
I haven't finished this yet, but please see #234

Thhethssmuz:
Ok, I've tried your branch and it seams to work for most cases. Though, and this might just be me trying to do something weird, but if you combine `noIntersperse` with a command parser:

```haskell
data Options = Options
  { foo  :: Bool
  , file :: String
  , args :: [String]
  } deriving Show

cmd :: ParserInfo Bool
cmd = info ( switch $ long "foo" ) idm

parser :: Parser Options
parser = Options
  <$> subparser ( command "cmd" cmd )
  <*> strArgument ( metavar "FILE" )
  <*> many ( strArgument $ metavar "ARGS..." )

main :: IO ()
main = do
  opts <- execParser $ info (helper <*> parser) noIntersperse
  print opts
```

[...]

PR #234, opened 2016-10-24
Fix noIntersperse behaviour, and add a new policy

HuwCampbell:
noInterperse was not allowing any options to occur
before arguments, and wasn't behaving as its namesake
in other parsing libraries.

Add a new policy (which is potentially useful for
applications wrapping other cli tools) allowing the
command (or subcommand) to collect all unknown options
and arguments as positionals to be passed to the
command line.

### DIFF
--- a/Options/Applicative/Types.hs
+++ b/Options/Applicative/Types.hs
@@ -84,8 +84,8 @@ data ParserInfo a = ParserInfo
   , infoHeader :: Chunk Doc   -- ^ header of the full parser description
   , infoFooter :: Chunk Doc   -- ^ footer of the full parser description
   , infoFailureCode :: Int    -- ^ exit code for a parser failure
-  , infoIntersperse :: Bool   -- ^ allow regular options and flags to occur
-                              -- after arguments (default: True)
+  , infoPolicy :: ArgPolicy   -- ^ allow regular options and flags to occur
+                              -- after arguments (default: InterspersePolicy)
   }
 
 instance Functor ParserInfo where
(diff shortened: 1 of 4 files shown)


SamTay:
Any idea when this well get merged? I too have built a CLI tool that forwards to another, and would love to save those characters / allow intuitive use. Right now `noIntersperse` breaks my executable..

HuwCampbell:
Plan is to get a point release out for documentation and version bounds before Christmas; then the 0.14 release with this (after I have figured out how to fix the lingering issue mentioned in #233) and #223.
So probably Jan.

SamTay:
Awesome, looking forward to it. Also thanks in general for your work, this package is wonderful.

Options/Applicative/Types.hs
```diff
@@ -319,12 +319,29 @@ type Args = [String]
...
-  deriving (Eq, Show)
+  = InterspersePolicy
+  -- ^ The default policy, options and arguments can
+  --   be interspersed.
+  --   A `--` option can be passed to ensure all following
+  --   commands are treated as arguments.
+  | NoInterspersePolicy
+  -- ^ Options must all come before arguments, once a
+  --   single option is parsed, all remaining arguments
+  --   are treated as positionals.
```
jystic: `s/once a single option is parsed/once a single argument is parsed/` ? (unless I misunderstand the behaviour)

jystic:
👍

erikd:
This looks useful! :+1:

PR #237, opened 2016-11-29
Topic/doc

HuwCampbell:
Putting this here as it's almost ready


README.md
```diff
@@ -1,32 +1,73 @@
...
+# optparse-applicative
 
 [![Continuous Integration status][status-png]][status]
 [![Hackage page (downloads and API reference)][hackage-png]][hackage]
 
+optparse-applicative is a haskell library for parsing options on the command line.
+
+Providing a powerful [applicative] interface to compose command line
+options and arguments, optparse-applicative provides all that is required for
+precise and concise CLI specifications for both simple and complex programs.
```
tmcgilchrist: I'd split this into two sentences, as it's too complex as is. Or even just make it a simpler introduction.
eg "optparse-applicative provides everything required ... "

README.md
```diff
@@ -1,32 +1,73 @@
...
+data Parser a
+
+instance Functor Parser
+instance Applicative Parser
+instance Alternative Parser
 ```
 
-Here is a simple example of an applicative option parser:
+A value of type `Parser a` represents a specification for a set of options,
+which, if parsed correctly, will yield a value of type `a`.
```
tmcgilchrist: perhaps swap `, if parsed correctly,` like "which will yield a value of type `a` if parsing succeeds`

README.md
```diff
@@ -39,102 +80,286 @@ data Sample = Sample
...
+basic combinators. In this example, hello is defined as an option
+with a `String` argument, while quiet is a boolean flag (called switch)
+and repeat gets parsed as an `Int` with help of the `Read` type class.
 
+The parser can be used like this:
+
+```haskell
 main :: IO ()
-main = execParser opts >>= greet
+main = greet =<< execParser opts
```
tmcgilchrist: Would a do block be more familiar / approachable for less experienced users?

README.md
```diff
@@ -39,102 +80,286 @@ data Sample = Sample
...
+         <> showDefault
+         <> value 1
+         <> metavar "INT" )
+```
 
-```haskell
-greet :: Sample -> IO ()
-greet (Sample h False n) = replicateM_ n $ putStrLn $ "Hello, " ++ h
-greet _ = return ()
+The parser is built using [applicative] style starting from a set of
```
tmcgilchrist: s/using/using an/

README.md
```diff
@@ -39,102 +80,286 @@ data Sample = Sample
...
+```
+
+Now `--file "foo.txt"` will be parsed as `FileInput "foo.txt"`, `--stdin` will
+be parsed as `StdInput`, but a command line containing both options, like
+
+    --file "foo.txt" --stdin
+
+will be rejected.
+
+Having `Applicative` and `Alternative` instances, optparse-applicative `Parser`s
```
tmcgilchrist: s/`Parser`s/Parsers or `Parser`'s

doesn't really make sense to have it quoted here

README.md
```diff
@@ -170,12 +395,12 @@ strOption
...
-name "o". See below for more information on the builder syntax and modifiers.
+`FILE` in the help text and documentation), default value "out.txt", a long
+name "output" and a short name "o".
 
-A regular option can return an object of any type, and takes a *reader*
+A regular `option` can return an object of any type, and takes a *reader*
 parameter which specifies how the argument should be parsed.  A common reader is
-`auto`, which assumes a `Read` instance for the return type and uses it to parse
+`auto`, which requires a `Read` instance for the return type and uses it to parse
 its argument. For example:
```
tmcgilchrist: Is there an example of running your own say Attoparsec parsers in this case?

README.md
```diff
@@ -193,31 +418,14 @@ output type. There's usually no need to add type annotations, however, because
...
-
-One can also use `ReadM` directly, using `str` to obtain the command line string,
-and `readerAbort` or `readerError` within the `ReadM` monad to exit with an
-error message.
+Further information on *readers* is available [below](#option-readers).
 
 ### Flags
 
-A **flag** is just like a regular option, but it doesn't take any arguments: it is
+A *flag* is just like a regular option, but it doesn't take any arguments: it is
```
tmcgilchrist: Again, not sure about the colon usage here.

README.md
```diff
@@ -337,47 +552,403 @@ main :: IO ()
...
+has been applied to the `Parser`.
+
+Authors can also use the preferences `showHelpOnError` or `showHelpOnEmpty` to show
+the help text on any parser failure or when a command is not complete and at the
+beginning of the parse of the main program or one of its subcommands respectively.
+
+Even if the help text is not shown for an error, a specific error message will
+be, indicating what's missing, or what was unable to be parsed.
+
+```
```
tmcgilchrist: Should be tagged as haskell

README.md
```diff
@@ -337,47 +552,403 @@ main :: IO ()
...
+string, and `readerAbort` or `readerError` within the `ReadM` monad to exit with an
+error message.
+
+One nice property of `eitherReader` is how well it composes with [attoparsec]
+parsers with
+
+```haskell
+import qualified Data.Attoparsec.Text as A
+attoReadM :: A.Parser a => ReadM a
+attoReadM p = eitherReader (A.parseOnly p . T.pack)
```
tmcgilchrist: Here it is :-)

README.md
```diff
@@ -1,32 +1,73 @@
...
 
-Here is a simple example of an applicative option parser:
+A value of type `Parser a` represents a specification for a set of options,
+which, if parsed correctly, will yield a value of type `a`.
+
+If you are familiar with parser combinator libraries like [parsec],
+[attoparsec], or the json parser [aeson] you will feel right at home with
+optparse-applicative.
+
+If not, don't worry: all you really need to learn are a few basic parsers, how
```
tmcgilchrist: Not sure the use of a colon is correct here. I'd just make it "If not, don't worry all you ..."

README.md
```diff
@@ -337,47 +552,403 @@ main :: IO ()
...
+  "--bar", is "--foo --bar" the option with value "--bar", or the default value with
+  the flag switched on? What if instead of a switch we had many positional string
+  arguments, is the first string the option's value or the first positional?
+
+  It is suggested to instead use the `Alternative` instance of `Parser` and create
+  a flag', an option, and a pure value for the default (with different names for the
+  flag and option).
+
+## How it works
+A Applicative `Parser` is essentially a heterogeneous list or tree of `Option`s,
```
tmcgilchrist: s/A/An

README.md
```diff
@@ -337,47 +552,403 @@ main :: IO ()
...
+  optArgs       <- many (argument str idm)
+  pure Options {..}
+```
+
+Here we've also used the *RecordWildCards* extension to make the parser
+specification cleaner. Compilation errors referring to `Monad` instances not
+being found are likely because the `Parser` specified can not be implemented
+entirely with `Applicative` (Note however, there were a few desugaring bugs
+regarding ApplicativeDo in GHC 8.0.1, function application with `($)` in
+particular may not work, and the `return` value instead be wrapped
```
tmcgilchrist: The example above uses `pure` rather than `return`? Slightly confusing perhaps

PR #241, opened 2017-02-09
Bump upper bound on process

bgamari:


HuwCampbell:
Cheers, just in time for #240

PR #244, opened 2017-02-21
Add suggestions for typos

### LINKED ISSUE #222: Subcommand suggestions
When one mis-types a subcommand. For example:

```
> git clonce
git: 'clonce' is not a git command. See 'git --help'.

Did you mean this?
    clone
```

A simple edit distance could be used.

Currently just a thought.

HuwCampbell:
Uses a Levenshtein distance to see if there's a suitable
candidate for suggestions.

Also fixes a subtle bug in bash completions, where argument
completers from deeper into the parser would have their
possibilities added to the completion.

Fixes #222

### DIFF
--- a/Options/Applicative/BashCompletion.hs
+++ b/Options/Applicative/BashCompletion.hs
@@ -40,16 +40,30 @@ bashCompletionQuery pinfo pprefs ws i _ = case runCompletion compl pprefs of
   Just (Right c)             -> run_completer c
   _                          -> return []
   where
-    list_options =
-        fmap concat
+    list_options
+      = fmap concat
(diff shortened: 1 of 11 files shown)


HuwCampbell:
@joeyh you've added something like this to git-annex I believe. Any thoughts?

Will this break your implementation?

joeyh:
git-annex contains code to handle typos the same as git does, but only for sub-commands. Since git-annex doesn't use optparse-applicative to parse the subcommand part of the command line, I don't anticipate this patch causing me trouble. And it will be nice to have typo suggestions for options.

PR #250, opened 2017-03-19
Fix Quick Start example in README.md

maxkwallace:
Without these changes, I saw:

```
$ runhaskell replace.hs 

replace.hs:35:28:
    Not in scope: ‘replicateM_’
    Perhaps you meant ‘replicate’ (imported from Prelude)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ parsers, and how to compose them as instances of `Applicative` and
 Here's a simple example of a parser.
 
 ```haskell
+import Control.Monad
 import Options.Applicative
 import Data.Semigroup ((<>))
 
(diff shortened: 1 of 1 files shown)


HuwCampbell:
G'day,

Thanks for this.

Just for style, can you change it to an explicit import (like the semigroups import) and remove the qualification at the call site?

HuwCampbell:
Or maybe just move the import to the code fence it's used in.

I don't want to trick people into thinking that they need to import `Control.Monad` to use the library.

HuwCampbell:
Have changed it to just use `replicate`. Thanks for the report.

maxkwallace:
@HuwCampbell thanks! Sorry for the slow reply, work has been busy recently :)

PR #251, opened 2017-03-21
Allow overriding the help info of a Parser

mgsloan:
This commit allows for specifying custom help info for a Parser.

The specific motivation for it is that it allows us to remove an ugly hack in stack and fix its bash completion.  In particular, we have a bunch of options that are enableable or disableable, like `--no-system-ghc` and `--system-ghc`.  Rather than list both, we have `--[no-]system-ghc` in the help list.  The current hack for this means that `--[no-]system-ghc` is an actual flag accepted by stack, and also breaks autocompletion! Argh!

Here's the stack PR that depends on this PR: https://github.com/commercialhaskell/stack/pull/3080

This will also be useful for things like `--PROG-options` and `--with-PROG`.  See, for example, this portion of the output of `cabal build --help`:

```
    --with-PROG=PATH    give the path to PROG
    --PROG-option=OPT   give an extra option to PROG (no need to quote options
                        containing spaces)
    --PROG-options=OPTS give extra options to PROG
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Adds `overrideHelp` and `overrideHelpDoc`, to allow custom help documentation.
+
 ## Version 0.13.2.0 (9 Mar 2017)
 
 - Updated dependency bounds.
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -168,6 +168,9 @@ searchParser f (BindP p k) = msum
   , case evalParser p of
       Nothing -> mzero
       Just aa -> searchParser f (k aa) ]
+searchParser f (HelpP hs p) = do
+  p' <- searchParser f p
+  return (HelpP hs p')
 
 searchOpt :: MonadP m => ParserPrefs -> OptWord -> Parser a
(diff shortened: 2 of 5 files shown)


mgsloan:
Will also help with https://github.com/commercialhaskell/stack/issues/3081

HuwCampbell:
I see. I'll have to think about this one though, it touches on and changes some central definitions which are central to parsing and evaluation, not just help text generation.

I would like to find a solution which doesn't affect the `Parser` type.

mgsloan:
I agree that it feels ugly to touch the Parser type.  Another approach, which I think may be even uglier would be to add the following:

1) A way to add an option without it being used for parsing, just for help.

2) A new variety of visibility that hides the option from the help docs but allows it in completions.

HuwCampbell:
I've a big priority to keep optparse clean, so I don't think I can merge this as is.

1) I don't think is a good idea either. 

2) May be possible

3) I could also see that one could maybe add a builder with type `Doc -> Doc` which is applied to the help description. People could also use it to colour some options maybe.

No guarantees though.

HuwCampbell:
I have edited the comment above, as it may have come across as rude first time round.

PR #253, opened 2017-03-23
Add a new modifier for editing an option description

HuwCampbell:
@mgsloan

Here's the general idea I mentioned in #251 

It touches less and I think gives a nicer level of expressiveness. I tried
`style (bold . underline)`
and it looked pretty good.

Your problem would be solved by adding to one of the flags
`style (const $ text "--[no]-whatever")` and hiding the other.

I'm not sure about the exact type or name of the builder yet, `Doc -> Doc` might actually do, as I don't think deleting or adding after it's been removed for other reasons is a good idea.

### DIFF
--- a/Options/Applicative/Builder.hs
+++ b/Options/Applicative/Builder.hs
@@ -41,6 +41,7 @@ module Options.Applicative.Builder (
   ParseError(..),
   hidden,
   internal,
+  style,
   command,
   commandGroup,
   completeWith,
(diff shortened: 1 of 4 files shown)


HuwCampbell:
@mgsloan  Does this make sense as to how it can solve your problem?

mgsloan:
Hmm, yes, that should solve part of the problem, thanks for working on it!

The other part of the problem is how to hide the flag from both full and brief docs while allowing it in bash completion.  This could be a new constructor for `OptVisibility`, but not sure what to call it.

HuwCampbell:
Cool. I might punt on the visibility part until ~0.15, as it's a little fiddly and not an issue for most users.

PR #254, opened 2017-03-23
Add an error message for options which require arguments

### LINKED ISSUE #249: Default error message for a missing argument to an option is an empty string
It's not very helpful.

HuwCampbell:
I believe this is the last situation in optparse-applicative where we don't emit an informative error message (by default).

Closes #249

### DIFF
--- a/Options/Applicative/Types.hs
+++ b/Options/Applicative/Types.hs
@@ -63,6 +63,7 @@ data ParseError
   | ShowHelpText
   | UnknownError
   | MissingError IsCmdStart SomeParser
+  | ExpectsArgError String
   | UnexpectedError String SomeParser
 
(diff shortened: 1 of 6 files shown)


Options/Applicative/Types.hs
```diff
@@ -63,6 +63,7 @@ data ParseError
   | ShowHelpText
   | UnknownError
   | MissingError IsCmdStart SomeParser
+  | ExpectsArgError (Maybe String)
```
HuwCampbell: This works, but there may be a better way to get the invariants right.

I could imaging a `ExpectsArg (String -> ParseError)` for instance, which would give the user more flexibility, but push the error message to the builder instead of renderer. One might change the builder to `String -> ParseError`, but that would break backwards compatibility.

PR #256, opened 2017-03-30
zsh and fish shell rich completions

HuwCampbell:
Adds zsh shell and makes fish shell's completion quite a lot nicer
with descriptions coming through for both

Here's fish
screen shot 2017-03-30 at 12 35 50 pm

And zsh
screen shot 2017-03-30 at 1 35 30 pm

Thanks for the start @Drezil


alem0lars-yoroi:
Can this pull request be merged?

HuwCampbell:
Don't worry, I'll merge it when I'm ready.
In the background I've been reaching out to knowledgeable zsh users, asking them to see if they can break it.
I also may permit a little customisation with regards to how much description is show for each command before cutting them off with an ellipsis. For zsh I'm now showing them on separate lines, like this
screen shot 2017-04-04 at 10 18 07 pm
So it probably should show more information for commands .

PR #259, opened 2017-04-24
Update Common.hs to expose stepParser

JonCoens:
stepParser is useful in building an incremental parser. Without exposing this, one needs to copy nearly all the contents of Options.Applicative.Internal. This change would save me from copying hundreds of LoC and save me from depending on internals that are likely to change.

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -47,7 +47,8 @@ module Options.Applicative.Common (
   -- * Low-level utilities
   mapParser,
   treeMapParser,
-  optionNames
+  optionNames,
+  stepParser
   ) where
 
 import Control.Applicative


HuwCampbell:
I'm not fundamentally opposed to exporting more functions, but this is a rather internal function, and it would be very strange to make it part of the public API. If I was going to export it, I would probably prefer to put it inside a module marked internal.

What are you trying to do that would require this? I may be able to help you find an alternative way.

JonCoens:
Thanks for responding so quickly!

**High level:** I want a parser that consumes all arguments given and returns a successfully parsed object as well as the set of arguments it knew nothing about. i.e. `:: ParserInfo a -> IO (a, [String])`

**More detail:**
I'm running in a mixed-language environment (Haskell + C++) where command-line flags are defined and processed in different places. Instead of users needing to remember which is a Haskell flag or a c++ flag, I'd rather have one of the parsers be smart enough to consume only what's required and give back the remaining set. This boils down to not failing when coming across an unrecognized flag. Having a required flag missing or giving a flag the wrong type are still grounds for quitting.

I'd be happy to open an Issue to track making this happen if that's easier. By exposing `stepParser` I can mimic parts of `execParser` that do what I want. I attempted making the changes I need (just a 2 line change to `runParser`), but it has some cascade effects into the test framework that I haven't been able to figure out.

HuwCampbell:
Yep, I've seen things like this before and recently made changes to support it more easily. Have you considered or seen the [ForwardOptions](https://github.com/pcapriotti/optparse-applicative/blob/master/Options/Applicative/Types.hs#L353) policy?

This will run the parser, but allow any flags and options unknown to the haskell parser to be collected into a positional arguments reader. Your c program could then run over those afterwards.

JonCoens:
Ah, I didn't realize that flag existed. I've been working via stackage, which doesn't have `ForwardOptions` in it. From the description that sounds like what I'd be looking for!

HuwCampbell:
It's slated in for the 0.14 release. Which should come next month.

HuwCampbell:
I'll close this then.

JonCoens:
Thank you!

PR #261, opened 2017-05-02
Fixed typo in README

bkase:
The documentation refered to a long option "hello", but in the example
the option is named "target".


HuwCampbell:
Thanks. I think it would be better flipped the other way though (with long option "hello"), so the example is the same as those above.

bkase:
@HuwCampbell cool, that makes sense -- let me know if it looks good now (I flipped it)

HuwCampbell:
Yep.

Thanks!

PR #266, opened 2017-05-26
Make `optparse-applicative` forward-`-Wcompat`ible

hvr:
See also
https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#base-4.9.0.0

### DIFF
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -51,10 +51,11 @@ source-repository head
   location: https://github.com/pcapriotti/optparse-applicative.git
 
 library
-  if impl(ghc >= 8)
-    ghc-options:       -Wall -Wno-redundant-constraints -Wcompat
-  else
-    ghc-options:       -Wall
+  ghc-options:         -Wall
+
+  -- See https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#base-4.9.0.0
+  if impl(ghc >= 8.0)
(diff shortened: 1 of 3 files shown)


optparse-applicative.cabal
```diff
@@ -51,10 +51,11 @@ source-repository head
...
 library
-  if impl(ghc >= 8)
-    ghc-options:       -Wall -Wno-redundant-constraints -Wcompat
-  else
-    ghc-options:       -Wall
+  ghc-options:         -Wall
+
+  -- See https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#base-4.9.0.0
+  if impl(ghc >= 8.0)
+    ghc-options:  -Wno-redundant-constraints -Wcompat -Wnoncanonical-monad-instances -Wnoncanonical-monadfail-instances
```
HuwCampbell: Is it ok to have two active `ghc-options` in scope like this? Will `Wall` still be on for ghc 8?

HuwCampbell:
Looks good, I just managed to get compat working for the remaining semigroup classes a few weeks ago, so I'm happy to future proof further.

hvr:
@HuwCampbell I actually had a patch sitting here which included the semigroup classes... but I forgot to make it into a PR... so this PR is actually a stripped down version of that :-)

optparse-applicative.cabal
hvr: `ghc-options` are combined via a concatenation monoid. That generally applies for all cabal  fields which are list-ish (such as options or `other-modules` or `build-depends` etc).

HuwCampbell:
Thanks

PR #267, opened 2017-05-27
Back out eitherReader polymorphism

HuwCampbell:
In #223 I made `eitherReader` and `maybeReader` provide any
`IsString` to the passed in function, but I'm not sure it's a great idea
for backwards compatibility.

A somewhat common use case for `eitherReader` and `maybeReader`
is something like
```haskell
a = eitherReader $ \case
      "this" -> Right This
      "that" -> Right That
      _      -> Left "Must be this or that"
```

And there's not a lot we can do for type inference if overloaded
strings is on. If we were to release with this change, then cases
like this would be a bit more challenging.

### DIFF
--- a/Options/Applicative/Builder.hs
+++ b/Options/Applicative/Builder.hs
@@ -130,17 +130,18 @@ str = fromString <$> readerAsk
 -- easily with
 -- @
 -- import qualified Data.Attoparsec.Text as A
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Putting this here for comment before I release 0.14 in a few days.

PR #268, opened 2017-06-06
Explicitly export all functions and types, instead of the modules

HuwCampbell:
Although it's a bit more work, being explicit about exported functions makes the haddocks much nicer.

### DIFF
--- a/Options/Applicative.hs
+++ b/Options/Applicative.hs
@@ -43,12 +43,12 @@ module Options.Applicative (
 
   -- ** Parser builders
   --
-  -- | This module contains utility functions and combinators to create parsers
+  -- | This section contains utility functions and combinators to create parsers
   -- for individual options.
   --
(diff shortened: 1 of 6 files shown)


HuwCampbell:
This might be a nice time to remove deprecated functions and tidy up the main module exports as well.

We export a few functions which can't be used without other qualified imports (such as `runParserFully`) and don't make sense to be exported from the top.

Options/Applicative.hs
```diff
@@ -186,33 +192,30 @@ module Options.Applicative (
...
-
-  -- ** Low-level utilities
-  mapParser,
-  treeMapParser,
-  optionNames,
-  liftOpt,
-  showOption,
+  -- * Types
+  ParseError(..),
+  ParserHelp(..),
```
HuwCampbell: I've added `ParserHelp` to main module export list as it's required for `overFailure` to be useful.

PR #276, opened 2017-09-24
Fix type signature in eitherReader example

gxmas:
In the example for `eitherReader`, the type signature for `attoparsecReader` has `A.Parser a` as a type constraint where it should be a type.


HuwCampbell:
Ta.

HuwCampbell:
Ahh, I thought I had fixed that, but apparently I only found the readme, not that one.

PR #285, opened 2017-12-30
add @since 0.13.0.0 to Mod Semigroup instance doc

chris-martin:


HuwCampbell:
Seems legit. This one was a bit breaking, so it's good to document when it happened.

PR #289, opened 2018-02-26
Bump upper bound to support transformers-compat 0.6

gwils:


HuwCampbell:
Thanks George,

Yeah I've noticed this one already. I'm currently actually scoping out the opposite solution of just dropping this dependency.

I believe that the only compatibility gained is for GHC versions 7.6 and under using transformers 0.3. Not sure it's worth it anymore to be honest.

Happy to be convinced otherwise.

Huw

gwils:
I don't have very strong feelings either way. Both options are better than the current state of things though. So go ahead with removing the dependency.

HuwCampbell:
I'll just merge this one as is and deal with deprecating (very) old versions of GHC in the next major version bump.

gwils:
That sounds like a good plan to me. In the meantime, would you please make a release with this patch included?

HuwCampbell:
Way onto it #290

HuwCampbell:
Done

Issue #58, opened 2013-11-19
Optparse hanging infinitely & support for multiple usage of the same flag


wdanilo:
Hi!
Would it be possible to add support for multiple usage of the same flag?
I would love to allow users to use arguments like this `myprog --dump ast --dump va --dump test` and I would like to get array of String's as a result.

And here is what I was trying (what was natural to me, because I have some Parsec background) and it hanged optparse when printing help:

`[...] many (strOption (long "dump" <> value "" <> metavar "DUMP")) [...]`

Thank you.

danse:
i stumbled upon this same problem today. i tried a clumsy `option (many str)` and got a scary runtime loop. this seems like a noob trap, i propose to improve the README with this example, i'll put a quick patch together

danse:
actually the best solution would probably be a function like `manyStrOption = many StrOption`, where the haddock can explain that this is a simple alias. to me this seems the more findable way to avoid similar error to new users of the library, and make adoption easier ... i see that you kept the interface very dry and orthogonal though so you probably want to keep it so

danse:
it could be modified as in #295 but it's not very visible, maybe we could have some examples of this usage elsewhere, or point to a guide which goes through applicative parsing?

PR #295, opened 2018-03-20
elaborate about the use of alternative instance of parsers

danse:
see #58


HuwCampbell:
I think I cover this off pretty well in the arguments section

> creates an argument accepting any string. To accept an arbitrary number of arguments, combine the argument builder with either the many or some combinator:
> 
> some (argument str (metavar "FILES..."))

No?

danse:
Well in my case i didn't look under Arguments because i was trying to build an option. I think that a bit of redundancy in documenting this wouldn't hurt. Alternatively the topic could be explained in a small dedicated section when introducing the principles underpinning the lib, i had found it helpful when starting off

HuwCampbell:
I'll add some extra docs.

Issue #298, opened 2018-04-27
Make `showHelpOnError` show help text for subcommand not top-level command on error


sseefried:
Let's say I'm creating a utility called `my-tool`
Say I have a `ParserInfo` value defined as:

```Haskell
parserInfo = info (options <**> helper) fullDesc
  where
    options = hsubparser (cmd1 <> cmd2)
```
where `cmd1` and `cmd2` are subparsers  with their own helpers defined.

e.g.
```Haskell
cmd1 = command "cmd" $ info (cmd1Parser <**> helper) fullDesc
```

Let's say I execute the parser with `customExecParser (prefs showHelpOnError) parserInfo`

Now let's say that, on the command line, I type something like:

```
$ my-tool cmd1 --some-flag-not-in-cmd1-parser
```

This produces an error. Unfortunately, what I'm seeing in the error output is something like this:

```
Invalid option `--some-flag-not-in-cmd1-parser'

Usage: my-tool COMMAND

Available options:
  -h,--help                Show this help text

Available commands:
  cmd1                Some description for cmd1
  cmd2                Some description for cmd2
```

I am seeing the help for the whole command, but what I actually want to see is the help for `cmd1`.
Is there some way for me to get this behaviour?

HuwCampbell:
I'll first explain what's happening here.

The default behaviour for optparse is backtracking once a subcommand is completed. So if cmd1 is "complete", meaning it has no more mandatory parsers to fill, any unrecognised options will cause the sub-parser to exit, to be tried by the parent parser. In your case the parent also can't find it, so the help message is emitted there.

This allows things like global options to be passed in after the subcommands, or multiple subcommands to be specified.

So you can turn off backtracking if you don't need it, and you'll only see the cmd1 usage.

HuwCampbell:
I'll close this, but feel free to follow up.

sseefried:
Thanks @HuwCampbell. I'll try this out on Monday.

sseefried:
I wonder if we can put this in the documentation somewhere as it requires quite a bit of knowledge of how the generated parsers work to work this out.  Thoughts?

moll:
I noticed there's also `SubparserInline` in http://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative-Types.html#t:Backtracking that seems to lack a description.

HuwCampbell:
[...]

PR #300, opened 2018-04-30
Add documentation on 'noBacktrack' option

sseefried:
Subcommand parsers are often disjoint, meaning that if one fails
to parse after the command name then all others (having distinct
command names) will also fail to parse. When the 'noBacktrack'
preference is not set but 'showHelpOnError' is this leads to
the help for the top-level parser being shown instead of the
help for the failing subcommand.

A note has been added to the Haddock docs for the 'noBacktrack'
option to make this clear.


sseefried:
@HuwCampbell this should avoid a repeat of someone like me creating a (non-) issue like #298  in the future :-)

HuwCampbell:
I've added documentation for these types.

PR #303, opened 2018-05-12
tidy up dependencies

vmchale:
This removes a dependency on `transformers-compat` where unnecessary.


HuwCampbell:
Thanks for the PR.

I believe that this is against the spirit of `transformers-compat` though, whose usual use case is to be imported and do more or less nothing where it's not required to.

PR #306, opened 2018-06-05
Don't set parsers to unreachable after a Command

### LINKED ISSUE #305: Bash autocompletion doesn't work with multiple commands
Autocompletion doesn't work for each command (only for first one) when I have multiple different commands at the same time.

Here is recorded demo of behavior:

[asciicast](https://asciinema.org/a/JUtqO6HnoaSvue0OCwLMDecaG)

And here is minimal code:

```haskell
module Main where

import Data.Semigroup ((<>))
import Options.Applicative (Parser, ParserInfo, command, execParser, fullDesc, helper, info,
                            metavar, optional, progDesc, subparser)

main :: IO ()
main = execParser prsr >>= print

data Options = Options (Maybe ()) (Maybe ())
    deriving Show

fooP :: Parser ()
fooP = subparser $ mconcat
    [ metavar "foo [OPTIONS]"
    , command "foo" $ info (helper *> pure ()) (progDesc "Foo")
    ]

barP :: Parser ()
barP = subparser $ mconcat
    [ metavar "bar [OPTIONS]"
    , command "bar" $ info (helper *> pure ()) (progDesc "Bar")
    ]

optsP :: Parser Options
optsP = Options <$> optional barP <*> optional fooP

prsr :: ParserInfo Options
prsr = info (helper <*> optsP) $ fullDesc <> progDesc "Completion example"
```

HuwCampbell:
Commands aren't absorbing, and any value which isn't one of the
specified subcommands can be passed to a latter command or argument.

Fixes #305

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -274,7 +274,7 @@ treeMapParser g = simplify . go False False False g
       | otherwise
       = MultNode []
     go m d r f (MultP p1 p2) = MultNode [go m d r f p1, go m d r' f p2]
-      where r' = r || has_positional p1
+      where r' = r || hasArg p1
     go m d r f (AltP p1 p2) = AltNode [go m d' r f p1, go m d' r f p2]
       where d' = d || has_default p1 || has_default p2
     go _ d r f (BindP p k) =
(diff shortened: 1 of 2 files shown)


ChShersh:
@HuwCampbell Now in my example when I try to autocomplete either `foo` or `bar` exactly after executable name:

```shell
$ foo-bar f TAB  # works
$ foo-bar b TAB  # works
```

autocompletion works great :+1: But when I try to autocomplete second command after typing first one, autocompletion for second command doesn't work :disappointed_relieved: 

```shell
$ foo-bar foo b TAB  # doesn't work
$ foo-bar bar f TAB  # doesn't work
```

HuwCampbell:
Ok.

This is happening because of backtracking. Essentially there are two parsers which can be looked at when running this completion, the one you've entered, and the parent parser. The parser is currently looking only in the entered subcommand.

I'll have a think about it...

Issue #294, opened 2018-03-20
Displaying global options when listing options for a command.


merijn:
I often find myself having the same options shared across multiple/all commands (i.e. verbosity) and specify those as global options and this all works fine. However, these global options are only shown when someone runs `foo --help` and not when running `foo subcommand --help`. Personally I find it rather annoying to have to run a command/help twice to figure out all the options I can specify.

I would like some way to either 1) annotate a command so that it will repeat all global options when it's help is displayed or 2) annotate a (global) option so that it is repeated for all commands.

HuwCampbell:
I'll have a think about it. I understand the frustration for sure.

My current thoughts are around how backtracking and "global" parser types will be inconsistent with this style of API. I can think of a few ways.

I do think there are new functions which can be written to ask for the "global options" help, and combine them with the subcommand help before pretty printing. We just need to find the right API.

domenkozar:
Got an user that was confused about the current `subparser` --help split: https://github.com/cachix/cachix/issues/126#issuecomment-424042513

HuwCampbell:
I have a dirty prototype version this working now.

I think for subparser inline it very much makes sense to show this information; I'm not quite so sure about the other run methods; it's probably best to not in no backtrack mode.

It's also a bit dirty with the way `--help` is implemented. As that's usually a separate parser per subcommand, so it shows up multiple times.

HuwCampbell:
This is the basic idea https://github.com/pcapriotti/optparse-applicative/compare/topic/global-options?expand=1

HuwCampbell:
Currently looks like this

```
Usage: cabal configure [--enable-tests] [-f|--flags FLAGS]
  Prepare to build the package

Available options:
  --enable-tests           Enable compilation of test suites
  -f,--flags FLAGS         Enable the given flag
  -h,--help                Show this help text

Global options:
  -v,--verbose LEVEL       Set verbosity to LEVEL
  --version                Print version information
  -h,--help                Show this help text
```

domenkozar:
:tada:

merijn:
I'll try this this weekend :)

HuwCampbell:
Cool. I've just fixed the duplicate help issue as well.

HuwCampbell:
Addressed in #385

PR #318, opened 2018-09-19
Allow inlining of subparsers instead of independent execution

HuwCampbell:


domenkozar:
Does this also address https://github.com/pcapriotti/optparse-applicative/issues/294?

HuwCampbell:
At the moment no.

That said, as it'll be possible to mix them with this flag on, it might make sense to support it. Essentially, it requires pushing into a context IFF the inline path is activated *after* `searchArg` has run.

jkachmar:
Is there anything preventing this from being merged? I just tested this branch and it solves a UX pain point in an application I'm working on quite nicely.

matthew-piziak:
Also looking forward to this.

HuwCampbell:
I'll include it in the next version (after a bit more testing in real projects).

domenkozar:
It would be lovely to have this out :)

jkachmar:
We’ve been using it in a production application since October without any ill-effect, if that bit of anecdata counts for anything :)

PR #331, opened 2019-02-26
relaxed Monoid constraint to Semigroup for the Monoid Chunk instance

benjaminselfridge:
Since `Chunk`s are supposed to lift a semigroup to a monoid by adding an identity, I thought it made sense to relax the `Monoid` constraint on the type parameter accordingly.

### DIFF
--- a/Options/Applicative/Help/Chunk.hs
+++ b/Options/Applicative/Help/Chunk.hs
@@ -45,10 +45,10 @@ instance Monad Chunk where
   return = pure
   m >>= f = Chunk $ unChunk m >>= unChunk . f
 
-instance Monoid a => Semigroup (Chunk a) where
-  (<>) = chunked mappend
+instance Semigroup a => Semigroup (Chunk a) where
+  (<>) = chunked (<>)
 
-instance Monoid a => Monoid (Chunk a) where
+instance Semigroup a => Monoid (Chunk a) where
   mempty = Chunk Nothing
   mappend = (<>)
 
--- a/Options/Applicative/Help/Chunk.hs
+++ b/Options/Applicative/Help/Chunk.hs
@@ -69,7 +69,7 @@ chunked f (Chunk (Just x)) (Chunk (Just y)) = Chunk (Just (f x y))
 --
 -- > isEmpty . listToChunk = null
 -- > listToChunk = mconcat . fmap pure
-listToChunk :: Monoid a => [a] -> Chunk a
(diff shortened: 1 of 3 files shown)


benjaminselfridge:
I suspected there might be a GHC version issue; perhaps it doesn't make sense to change this until some more time passes.

HuwCampbell:
Originally it was like this because `Doc` in ansi-wl-pprint didn't have a Semigroup instance. It does now however since 0.6.8. If this goes through we'll need to update its bounds as well.

The reason you're getting that error is because Semigroup was not a superclass of Monoid until GHC 8.4. So the signatures for a few functions will need to be changed as well.

I plan on ripping out most of ansi-wl-pprint and replacing with pretty-printer in the next month, so I can probably cover this off while doing that if you don't feel like polishing this up.

benjaminselfridge:
If the only thing that needs to be changed is `ansi-wl-pprint`, I'll just do that now, on top of this change, so it can be one less thing on the list.

benjaminselfridge:
All errors have been fixed except one that is unavoidable if we are bumping up the bounds for `ansi-wl-pprint` -- newer versions of that package require `base>=4.5`, and so GHC 7.0 and 7.2 will no longer build it. I don't know if that's a problem or not.

Options/Applicative/Help/Chunk.hs
```diff
@@ -69,7 +69,7 @@ chunked f (Chunk (Just x)) (Chunk (Just y)) = Chunk (Just (f x y))
 --
 -- > isEmpty . listToChunk = null
 -- > listToChunk = mconcat . fmap pure
-listToChunk :: Monoid a => [a] -> Chunk a
+listToChunk :: (Semigroup a, Monoid a) => [a] -> Chunk a
 listToChunk [] = mempty
 listToChunk xs = pure (mconcat xs)
```
HuwCampbell: `sconcat` will remove the need for `semigroup`.

HuwCampbell:
I've tried pretty hard to maintain good backwards compatibility, but I think bumping base and dropping support for 7.0 and 7.2 is fine. After all, they're now 7 years old or more.

Thanks :)

benjaminselfridge:
OK -- I'll delete those lines in the travis config and you can merge if/when you feel like it. Thank you.

Options/Applicative/Help/Chunk.hs
benjaminselfridge: Just saw this, will resolve momentarily...

HuwCampbell:
ta

Issue #307, opened 2018-06-12
Bug: `optional` combinator produces incorrect `--help` docs


ntc2:
Consider the following simple program, which has two options, where it's legal to specify none or both of them, but not legal to specify just one of them:

``` haskell
import Options.Applicative

data Opts = Opts
  { b :: Maybe B
  } deriving Show

optsParser :: Parser Opts
optsParser = Opts
  <$> optional bParser

data B = B
  { b0 :: String
  , b1 :: String
  } deriving Show

bParser :: Parser B
bParser = B
  <$> strOption (long "b0")
  <*> strOption (long "b1")

main :: IO ()
main =
   print =<< execParser (info (helper <*> optsParser) fullDesc)
```

The help output incorrectly indicates that it's OK to specify just one of the options:

``` console
$ stack exec -- bug --help
Usage: bug [--b0 ARG] [--b1 ARG]
```

The correct usage would be `Usage: bug [--b0 ARG --b1 ARG]`, since just specifying one of the two options is not allowed. Indeed, running the program and specifying just one of the options is an error:

``` console
$ stack exec -- bug       
Opts {b = Nothing}
$ stack exec -- bug --b0 0 --b1 1
Opts {b = Just (B {b0 = "0", b1 = "1"})}
$ stack exec -- bug --b0 0       
Missing: --b1 ARG

Usage: bug [--b0 ARG] [--b1 ARG]
```

I believe this issue is closely related to Issue #220.

HuwCampbell:
Yep, this is a duplicate of #220.

The tree structure we currently use for pretty printing the command line simplifies what's optional and has multiplicity to the level of each option. It's this way because it's actually quite complex tracking this in the tree and getting the pretty printing right.

No excuse though, it should be fixed.

benjaminselfridge:
@HuwCampbell I'm working on a patch for this right now. There's a chance I won't finish it, but I wanted to note this here to in case you wanted to avoid duplicate efforts.

ntc2:
Thanks Ben!

PR #334, opened 2019-03-08
Bug fix for issue #307

benjaminselfridge:
This is a patch for issue #307 -- it may not be ready for merge yet, but I'm currently passing all the tests.

### DIFF
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -278,8 +278,10 @@ treeMapParser g = simplify . go False False False g
       MultNode [go m d r f p1, go m d r' f p2]
       where r' = r || hasArg p1
     go m d r f p@(AltP p1 p2) =
-      AltNode (has_default p && not (has_default p1 && has_default p2)) [go m d' r f p1, go m d' r f p2]
-      where d' = d || (has_default p && not (has_default p1 && has_default p2))
+      AltNode altNodeType [go m d' r f p1, go m d' r f p2]
+      where altNodeType | has_default p && not (has_default p1 && has_default p2) = AltDefault
+                        | otherwise = AltNoDefault
+            d' = d || altNodeType == AltDefault
     go _ d r f (BindP p k) =
       let go' = go True d r f p
       in case evalParser p of
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -279,9 +279,19 @@ treeMapParser g = simplify . go False False False g
       where r' = r || hasArg p1
     go m d r f p@(AltP p1 p2) =
       AltNode altNodeType [go m d' r f p1, go m d' r f p2]
-      where altNodeType | has_default p && not (has_default p1 && has_default p2) = AltDefault
-                        | otherwise = AltNoDefault
-            d' = d || altNodeType == AltDefault
+      where
+        -- The 'altNodeType' variable tracks whether or not this node of the
+        -- 'OptTree' ought to be displayed with brackets or not. Generally, we want
+        -- to put brackets around it when the parser has optional arguments, but if
+        -- *both* of its children also have optional arguments, then we don't put
+        -- brackets around the top-level because that would be redundant.
+        altNodeType | has_default p && not (has_default p1 && has_default p2) = AltDefault
+                    | otherwise = AltNoDefault
+        -- The 'd' variable tracks whether the option nodes at the leaves have
+        -- optional arguments so that when we want to hide optional arguments, we
+        -- have the information needed to do that. An option can be considered
+        -- optional in this sense if any of its parents were optional.
+        d' = d || altNodeType == AltDefault
     go _ d r f (BindP p k) =
       let go' = go True d r f p
       in case evalParser p of
--- a/Options/Applicative/Common.hs
+++ b/Options/Applicative/Common.hs
@@ -251,7 +251,7 @@ mapParser f = flatten . treeMapParser f
   where
     flatten (Leaf x) = [x]
     flatten (MultNode xs) = xs >>= flatten
-    flatten (AltNode xs) = xs >>= flatten
+    flatten (AltNode _ xs) = xs >>= flatten
 
 -- | Like 'mapParser', but collect the results in a tree structure.
 treeMapParser :: (forall x . OptHelpInfo -> Option x -> b)
@@ -262,7 +262,9 @@ treeMapParser g = simplify . go False False False g
     has_default :: Parser a -> Bool
     has_default p = isJust (evalParser p)
 
-    go :: Bool -> Bool -> Bool
+    go :: Bool
+       -> Bool
+       -> Bool
        -> (forall x . OptHelpInfo -> Option x -> b)
        -> Parser a
        -> OptTree b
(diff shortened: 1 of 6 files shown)


benjaminselfridge:
@HuwCampbell I'd love it if you wanted to take a look at this when you have the time.

HuwCampbell:
Elegantly done.

Would you mind cherry picking the two relevant commits onto the master branch as a first step? I'll leave inline comments.

Options/Applicative/Types.hs
```diff
@@ -384,7 +384,9 @@ data OptHelpInfo = OptHelpInfo
 data OptTree a
   = Leaf a
   | MultNode [OptTree a]
-  | AltNode [OptTree a]
+  | AltNode Bool [OptTree a]
+  -- ^ The 'Bool' argument indicates whether this node should be displayed with
```
HuwCampbell: You would probably be better off making a type for this instead of using Bool

Options/Applicative/Common.hs
```diff
@@ -272,10 +274,12 @@ treeMapParser g = simplify . go False False False g
...
       = MultNode []
-    go m d r f (MultP p1 p2) = MultNode [go m d r f p1, go m d r' f p2]
+    go m d r f (MultP p1 p2) =
+      MultNode [go m d r f p1, go m d r' f p2]
       where r' = r || hasArg p1
-    go m d r f (AltP p1 p2) = AltNode [go m d' r f p1, go m d' r f p2]
-      where d' = d || has_default p1 || has_default p2
+    go m d r f p@(AltP p1 p2) =
+      AltNode (has_default p && not (has_default p1 && has_default p2)) [go m d' r f p1, go m d' r f p2]
+      where d' = d || (has_default p && not (has_default p1 && has_default p2))
```
HuwCampbell: This logic is duplicated and getting relatively complex.
Please use a let binding and add an explanatory comment.

tests/test.hs
```diff
@@ -146,6 +146,19 @@ prop_alt_help = once $
...
+  let p :: Parser (Maybe (String, String))
+      p = optional ((,)
+                    <$> strOption ( long "a"
+                                    <> metavar "A"
+                                    <> help "value a" )
+                    <*> strOption ( long "b"
+                                    <> metavar "B"
+                                    <> help "value b" ) )
+      i = info (p <**> helper) idm
+  in checkHelpText "optional" i ["--help"]
```
HuwCampbell: Can you add the example from #307 showing nested optional inputs as well?

I've run it, it works, but it would be good to have it addressed in the tests too.

HuwCampbell: Sorry, I mean #220.

Options/Applicative/Help/Core.hs
```diff
@@ -59,12 +59,13 @@ optDesc pprefs style info opt =
...
         = mappend chunk suffix
-        | hinfoDefault info
-        = mappend (fmap brackets chunk) suffix
+        --  | hinfoDefault info
+        -- = mappend chunk suffix
         | null (drop 1 descs)
         = mappend chunk suffix
         | otherwise
-        = mappend (fmap parens chunk) suffix
+        = mappend chunk suffix
```
HuwCampbell: These 3 cases have the same result logic.
Please double check, and remove the guarded expressions if you're sure.

HuwCampbell:
Looks promising, but there seems to be dropped parenthesis when multiple names are used for a parser.

```
let p = flag' True (short 'A' <> long "all") 
main = execParser (info (helper <*> p) fullDesc) >> pure ()
:main --help
Usage: <interactive> (-A|--all)
```
vs with this change
```
:main --help
Usage: <interactive> -A|--all
```

Options/Applicative/Help/Core.hs
```diff
@@ -59,12 +59,13 @@ optDesc pprefs style info opt =
...
         | isEmpty chunk || not (descSurround style)
         = mappend chunk suffix
-        | hinfoDefault info
-        = mappend (fmap brackets chunk) suffix
+        --  | hinfoDefault info
+        -- = mappend chunk suffix
         | null (drop 1 descs)
         = mappend chunk suffix
         | otherwise
-        = mappend (fmap parens chunk) suffix
```
HuwCampbell: This set of parenthesis was important when there's more than one name

HuwCampbell:
Overall, I think this is a great improvement. We'll need to fix up the regression I've found, and tidy it up a bit.

Options/Applicative/Help/Core.hs
HuwCampbell:
```haskell
      render chunk
        | not show_opt
        = mempty
        | not (hinfoDefault info) && length descs > 1
        = fmap parens chunk <> suffix
        | otherwise
        = chunk <> suffix
```

Seems to do the trick

benjaminselfridge:
I'll take a look at all this today. 

> Elegantly done.
> 
> Would you mind cherry picking the two relevant commits onto the master branch as a first step? I'll leave inline comments.

This is the first time I've run into a situation where I had my own fork, which got merged back to master, and then I did more development on that same fork. I'm not exactly sure how to cherry-pick but I can try and figure it out.

benjaminselfridge:
Ok -- I'm going to create a different pull request with the cherry-picked changes and without all the other junk. Sorry about that. I'm going to leave this one open so it's easy to access all the comments you've written here. 

Let me know if you'd prefer me to do it another way -- sorry for the confusion.

benjaminselfridge:
I believe I did the cherry-picking thing properly. I'll address the other issues shortly.

Options/Applicative/Types.hs
benjaminselfridge: Done

Options/Applicative/Common.hs
benjaminselfridge: Done

tests/test.hs
benjaminselfridge: Done

Options/Applicative/Help/Core.hs
benjaminselfridge: Incorporated changes you've suggested below

Options/Applicative/Help/Core.hs
benjaminselfridge: Done

benjaminselfridge:
OK, I've addressed all your comments here. Unless there's anything else, I believe this is ready for merge.

benjaminselfridge:
OK, I've addressed all your comments above, and I think this is ready for merge (unless there is anything else).

HuwCampbell:
Thank you for this, I think it's great. 

I've found a few small issues with the rendering of parentheses. I'll pull these commits into a branch this evening (time permitting) and see if I can sort them out.

I believe that returning just `Chunk Doc` from `optDesc` isn't flexible enough, now that the rendering of brackets and parentheses is handled in fold_tree.

benjaminselfridge:
Excellent! Thanks for saying that.

Yeah, I figured there might be some details I didn't quite get right. Look forward to seeing the final merge. Thanks for checking it out so quickly.

HuwCampbell:
I believe I have a working solution, and have added a few extra tests which show some of the issues I found.

Unfortunately it still needs a refactor. It's a bit unreadable and hard to maintain at the moment.

HuwCampbell:
See https://github.com/pcapriotti/optparse-applicative/compare/topic/rewire-tree?expand=1

HuwCampbell:
Subsuming into a new branch

Issue #327, opened 2019-01-21
Equality sign in help


googleson78:
Is it possible to have a flag which modifies how long options are printed.
More specifically I'm talking about this:

Currently if you have a long option `foo` with an argument `ARG1` it will display as:
```
--foo ARG1
```
in the `--help`. But both `--foo arg1` and `--foo=ARG` parse correctly. Furthermore in most linux (unix?) utilities long options with arguments are set with `=` (in my experience at least). Is it possible to add a modifier/parser pref which will instead display `foo` as:
```
--foo=ARG1
```
in `--help`?

HuwCampbell:
This seems reasonable.

HuwCampbell:
So I've had a bit more of a think about this one. While it might be plausible to implement this, I think it may interact strangely when we have multiple names for an option.

For example
```
(-f|--foo ARG)
```
is fine, and both `-f ARG` and `--foo ARG` work. But `-f=ARG` isn't ok. So I don't think we should say it is.

benjaminselfridge:
I ran into the same issue when implementing this feature -- what's the appropriate way to handle multiple names, when some are short and some are long? This snippet from `man ls` may be a guide:

```
       -T, --tabsize=COLS
              assume tab stops at each COLS instead of 8
```

So `man` sometimes lists multiple names for options by putting the equals sign after the long one and not the short one. I think users can be expected to know what this means. I think writing

```
(-f|--foo=ARG)
```

in the usage string is probably clear to most people as well, although perhaps a bit less so. I think it would be worth experimenting with.

benjaminselfridge:
Experimented with this tonight -- see #336

HuwCampbell:
Done and part of 0.15.0.

Thanks @benjaminselfridge and @googleson78

PR #336, opened 2019-03-17
potential feature addition #327 -- optional equality sign in help

benjaminselfridge:
Added a `prefHelpLongEquals` preference modifier that displays equals signs for long options rather than spaces. Does not affect the behavior of the parser; it's purely a cosmetic alteration in terms of how the help and usage info are rendered. Have not yet tested this thoroughly, but it seems to look fine for all the obvious cases.

I did this in the most sensible way I could think of -- basically, all the long names are displayed with equals signs even if there are multiple long names, so `-l|--long1=|--long2=ARG` is how it would be displayed if you had one short and two long, for example. The way we detect whether to use an equals sign is by checking the `ParserPrefs` argument along with ensuring that the option actually takes a value (which, in turn, is determined by checking whether the metavariable is a `Chunk Just _` or a `Chunk Nothing`).

### DIFF
--- a/Options/Applicative.hs
+++ b/Options/Applicative.hs
@@ -195,7 +195,7 @@ module Options.Applicative (
   noBacktrack,
   subparserInline,
   columns,
-  longEquals,
+  helpLongEquals,
   defaultPrefs,
 
   -- * Completions
--- a/Options/Applicative.hs
+++ b/Options/Applicative.hs
@@ -195,6 +195,7 @@ module Options.Applicative (
   noBacktrack,
   subparserInline,
   columns,
+  longEquals,
   defaultPrefs,
 
   -- * Completions
--- a/Options/Applicative/Builder.hs
+++ b/Options/Applicative/Builder.hs
@@ -85,6 +85,7 @@ module Options.Applicative.Builder (
   noBacktrack,
   subparserInline,
   columns,
+  longEquals,
   prefs,
   defaultPrefs,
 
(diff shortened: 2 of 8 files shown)


benjaminselfridge:
@HuwCampbell I could use this myself, so I'd love to see it in the next release if it makes sense to you.

HuwCampbell:
Cheers.

I don't really like the current rendering when there are multiple long options, so I'll look at editing this.

benjaminselfridge:
I don't really like the rending either, but there has to be some solution.

One thing to keep in mind is that it's pretty rare (in my experience anyway) for an application to have multiple long names for the same functionality, so this is a relatively rare scenario to begin with. The benefit of allowing users of the library to display the `=` sign (which is ubiquitous when it comes to long names for options) greatly outweighs whatever ugliness would be incurred in the multiple-long-name scenario.

That's my two cents -- it's certainly worth thinking about a nicer way to render the multiple long names, but I do think this is a really nice feature that would improve the library overall, regardless of how that corner case ends up getting solved.

benjaminselfridge:
Another thought.

In the scenario where we have, say, two long options and one short one, my solution ends up looking like this:
```
Usage: op (-b|--with-bar=|--withb=ARG)

Available options:
  -b,--with-bar=,--withb=ARG
                           bar value
```
What about something like this instead:
```
Usage: op (-b ARG|--with-bar=ARG|--withb=ARG)

Available options:
  -b ARG,--with-bar=ARG,--withb=ARG
                           bar value
```
That is, explicitly listing how you specify the argument after each and every option rather than doing it only at the end?

HuwCampbell:
Subsumed into #347

PR #339, opened 2019-05-01
Support GHC 8.8

fumieval:
Pardon an automatically-posted pull request; this is a part of [Operation Vanguard](https://github.com/haskell-vanguard/haskell-vanguard). Note that this does not guarantee that tests and benchmarks are buildable.


HuwCampbell:
This looks just fine, I was wondering if you would be interested in providing a test to ensure there won't be any regressions?

HuwCampbell:
We've covered this off on another PR. Thank you.

PR #341, opened 2019-05-06
Update for GHC 8.8.1

vmchale:
This updates `optparse-applicative` for GHC 8.8.1. Doing so without CPP in the main library requires removing a `fail` method from the `Monad` class.


HuwCampbell:
Thanks for this.

It appears to be a duplicate of #339 of though. That one also appears to have done more work along the lines of backwards compatibility, which is very important for optparse.

PR #342, opened 2019-05-10
Allow semigroups-0.19 and fix Monad(fail)

phadej:
- semigroups-0.19 doesn't break anything in optparse-applicative
- fail isn't a method of Monad since base-4.13 (GHC-8.8)
  - Also the warning is removed, as it says
    -Wnoncanonical-monadfail-instances is deprecated: fail is no longer a method of Monad

This is superset of https://github.com/pcapriotti/optparse-applicative/pull/339

I'd also recommend using [`haskell-ci`](https://github.com/haskell-CI/haskell-ci) for generating .travis.yml; I'll do a separate PR with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
  - GHCVER=8.2.1
  - GHCVER=8.4.1
  - GHCVER=8.6.3
+ - GHCVER=8.8.1
  - GHCVER=head
 
 matrix:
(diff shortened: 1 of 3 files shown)


phadej:
I went ahead and made a revision to allow `semigroups-0.19` https://hackage.haskell.org/package/optparse-applicative-0.14.3.0/revisions/

PR #347, opened 2019-06-08
Topic/long equals

HuwCampbell:
@benjaminselfridge 

I've adjusted your PR to what I think is a better user interface. See the tests for appreciable differences. I'd appreciate your thoughts.

Cheers, Huw

### DIFF
--- a/src/Options/Applicative/Builder.hs
+++ b/src/Options/Applicative/Builder.hs
@@ -85,6 +85,7 @@ module Options.Applicative.Builder (
   noBacktrack,
   subparserInline,
   columns,
+  helpLongEquals,
   prefs,
   defaultPrefs,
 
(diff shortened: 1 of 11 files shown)


tests/test.hs
```diff
@@ -202,6 +202,28 @@ prop_nested_optional_help = once $
...
+prop_short_no_separation :: Property
+prop_short_no_separation = once $
+  let p :: Parser String
+      p = option auto (   short 'i'
+                       <> help "integer value")
+      i = info (p <**> helper) fullDesc
+      result = execParserPure (prefs helpLongEquals) i ["--help"]
+  in assertError result $ \failure ->
+    let text = head . lines . fst $ renderFailure failure "test"
+    in  "Usage: test (-iARG)" === text
```
HuwCampbell: Notice that short args don't get an extra space.

tests/long_equals.err.txt
```diff
@@ -0,0 +1,6 @@
+Usage: long_equals (-i|-j|--intval|--intval2=ARG)
```
HuwCampbell: And we only do the equality sign once.

src/Options/Applicative/Builder.hs
```diff
@@ -500,6 +501,11 @@ subparserInline = PrefsMod $ \p -> p { prefBacktrack = SubparserInline }
 columns :: Int -> PrefsMod
 columns cols = PrefsMod $ \p -> p { prefColumns = cols }
 
+-- | Show equals sign, rather than space, in usage and help text for options with
+-- long names.
+helpLongEquals :: PrefsMod
+helpLongEquals = PrefsMod $ \p -> p { prefHelpLongEquals = True }
```
HuwCampbell: I want to change the name of this builder.

tests/test.hs
HuwCampbell: I've reverted this one.

PR #354, opened 2019-07-17
Improve readability of help text

HuwCampbell:
Union types, which are surrounded by either brackets or
parenthesis, shouldn't dangle both sides in the middle
of a line.

If we see a union, we should either fit it on the current
line, or start a new line.

@domenkozar


HuwCampbell:
It's a bit subtle.

Issue #362, opened 2019-10-29
Indicate that a group of options can be repeated.


crocket:
```
{-# LANGUAGE LambdaCase #-}
import qualified Options.Applicative as O
import Control.Applicative (optional, some)
import qualified Sound.ALSA.Mixer as AM

type Mixer = String
type Control = String
type Index = AM.CUInt
type Label = String
data DisplayType = Switch | Db
data DisplayItem = Item DisplayType Mixer Control (Maybe Index) Label

indexP :: O.Parser (Maybe Index)
indexP = optional $ O.option O.auto ( O.long "index"
                                      <> O.short 'i'
                                      <> O.metavar "INDEX"
                                      <> O.help "index for control" )
parser :: O.Parser [DisplayItem]
parser = some $ Item
  <$> O.option (O.maybeReader $ \case
                   "switch" -> Just Switch
                   "db" -> Just Db
                   _ -> Nothing)
  ( O.long "type"
    <> O.short 't'
    <> O.metavar "TYPE"
    <> O.help "display type" )
  <*> O.strOption ( O.long "mixer"
                    <> O.short 'm'
                    <> O.metavar "MIXER"
                    <> O.help "mixer for switch" )
  <*> O.strOption ( O.long "control"
                    <> O.short 'c'
                    <> O.metavar "CONTROL"
                    <> O.help "mixer control" )
  <*> indexP
  <*> O.strOption ( O.long "label"
                    <> O.short 'l'
                    <> O.metavar "LABEL"
                    <> O.help "label" )

main :: IO ()
main = do
  items <- O.execParser $ O.info (parser O.<**> O.helper)
    ( O.progDesc "display switches and decibels for i3bar")
```

In ghci REPL

```
:main -h
Usage: <interactive> (-t|--type TYPE) (-m|--mixer MIXER) (-c|--control CONTROL)
                     [-i|--index INDEX] (-l|--label LABEL)
  display switches and decibels for i3bar

Available options:
  -t,--type TYPE           display type
  -m,--mixer MIXER         mixer for switch
  -c,--control CONTROL     mixer control
  -i,--index INDEX         index for control
  -l,--label LABEL         label
  -h,--help                Show this help text
*** Exception: ExitSuccess
```

There is no indication that a group of options can be repeated.

HuwCampbell:
[...]

PR #363, opened 2019-11-06
Topic/multi suffix generalisation

HuwCampbell:
Allows suffixes for multi arguments to be applied to groups of arguments, addressing #362 .

As an indication, things like this ` (-a|-b)...` are now possible.

This one's a bit subtle in its implementation.

@crocket

### DIFF
--- a/src/Options/Applicative/BashCompletion.hs
+++ b/src/Options/Applicative/BashCompletion.hs
@@ -91,7 +91,7 @@ bashCompletionQuery pinfo pprefs richness ws i _ = case runCompletion compl ppre
     --
     -- For options and flags, ensure that the user
     -- hasn't disabled them with `--`.
-    opt_completions argPolicy hinfo opt = case optMain opt of
+    opt_completions argPolicy reachability opt = case optMain opt of
       OptReader ns _ _
          | argPolicy /= AllPositionals
         -> return . add_opt_help opt $ show_names ns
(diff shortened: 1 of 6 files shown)


HuwCampbell:
@phadej These test failures seem spurious. Any ideas what's going on?

phadej:
The hlint errors are definitely not spurious:

```
0.99s$ if [ $HCNUMVER -eq 80605 ] ; then (cd optparse-applicative-* && hlint -h ${TOP}/.hlint.yaml src) ; fi
src/Options/Applicative/Common.hs:290:29: Warning: Redundant bracket
Found:
  (go')
Perhaps:
  go'
src/Options/Applicative/Help/Core.hs:64:11: Suggestion: Use null
Found:
  length names == 0
Perhaps:
  null names
Note: increases laziness
src/Options/Applicative/Help/Core.hs:108:7: Suggestion: Redundant bracket
Found:
  (foldTree pprefs style) .
    mfilterOptional . treeMapParser (optDesc pprefs style)
Perhaps:
  foldTree pprefs style .
    mfilterOptional . treeMapParser (optDesc pprefs style)
```

phadej:
Searching (ctrl+f) for `exited with 1` helps find failing parts

HuwCampbell:
Ahh thank you. I see. I was looking for test failures.

I think these are a bit iffy in terms of "is there something which could lead to a run time issue"; and was surprised they only turned up for a specific GHC version.

I will fix these, but I don't think hlint warnings to errors are really necessary.

phadej:
You can remove HLint checks from Ci completely. If CI builds don’t fail, hardly anyone scans through the build logs.

HLint is run on one job because a) installing HLint takes some time b) it would give the same results anyway.

PR #371, opened 2020-02-04
Fixed warning and added support for stack

flip111:


HuwCampbell:
Thank you for putting in some work for optparse-applicative. Unfortunately however this breaks builds for GHC 8.2 and earlier.

flip111:
What is the minimum GHC version you would like to support ?

HuwCampbell:
As many as is feasible and not too painful.

We support back to 7.0 at the moment, which is about a decade old. I would be relatively comfortable only supporting 7.8 and above with a good reason.

While it's not a big deal though, I'm happy to keep supporting quite old versions.

HuwCampbell:
Ta

PR #382, opened 2020-05-01
Add qualified some1

HuwCampbell:
Because this does clash, a qualified version might be appropriate and easy.

### DIFF
--- a/src/Options/Applicative/NonEmpty.hs
+++ b/src/Options/Applicative/NonEmpty.hs
@@ -1,0 +1,17 @@
+{-# LANGUAGE CPP, Rank2Types, ExistentialQuantification #-}
+module Options.Applicative.NonEmpty (
+  some1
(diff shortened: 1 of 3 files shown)


src/Options/Applicative/NonEmpty.hs
```diff
@@ -0,0 +1,17 @@
...
+import Data.List.NonEmpty (NonEmpty (..))
+
+import Options.Applicative.Types
+import Control.Applicative
+import Prelude
+
+-- | Sequences an action one or more times.
+--
+--   Functionally identical to 'Data.List.NonEmpty.some1',
+--   but is preferred as it gives a nicer help text.
```
andreasabel, line 15: This haddockumentation should have a `@since` attribute.
https://haskell-haddock.readthedocs.io/en/latest/markup.html#since

Issue #359, opened 2019-09-23
Subcommand help text lacks required parent command arguments


symbiont-joseph-kachmar:
I'm currently using `optparse-applicative` to construct a CLI parser where the command tree has top-level commands that take required and optional inputs as well as subcommands that may make use of the inputs given to the parent command.

As far as I can tell, there's no way to have `optparse-applicative` generate help text and/or example usage that reflects the required input coming from the parent command parser.

---

Take the following program as an example, where I'm constructing an options parser for a parent command and a subcommand that allows subcommand's options to be passed inline with the parent command's.

https://gist.github.com/symbiont-joseph-kachmar/84a7e47fba5a7f1cc46c8603bd916f92

If I compile this to some executable `exe` and call `exe --help`, I get the following output:
```
Usage: exe (-a|--arg COMMAND_INPUT) [COMMAND]
  Optparse subcommand example...

Available options:
  -h,--help                Show this help text
  -a,--arg COMMAND_INPUT   Required command input

Available commands:
  sub                      Example subcommand
```
If I call `exe sub --help`, however, the "Available options" stanza makes no mention of the required `COMMAND_INPUT` option:
```
Usage: exe sub (-s|--sub-arg SUBCOMMAND_ARG)
  Example subcommand

Available options:
  -h,--help                Show this help text
  -s,--sub-arg SUBCOMMAND_ARG
                           Required subcommand input
```
Finally, if I call `exe sub`, the error response will correctly tell me which options were missing, but (again) these options won't be shown in the usage section:
```
Missing: (-a|--arg COMMAND_INPUT) (-s|--sub-arg SUBCOMMAND_ARG)

Usage: exe sub (-s|--sub-arg SUBCOMMAND_ARG)
  Example subcommand

Available options:
  -h,--help                Show this help text
  -s,--sub-arg SUBCOMMAND_ARG
                           Required subcommand input
```

---

I imagine that due to the nature of how `Parser`s are constructed, this type of thing would probably be difficult/impossible to support directly right now (I can't think of a way for the subcommand `Parser` to "know" about the required options of its parent).

What might be necessary to add support for this to the library; barring that, are there any patterns or approaches I could take to manually add support for this behavior?

If the relevant `Doc`s generating this text could be pulled out from the parent `Parser` I could pass that to the function constructing the subcommand `Parser` and I think that would be fine.

HuwCampbell:
Yeah... this is a bit tricky.

[...]

Issue #175, opened 2015-12-25
List detailed subparser documentation with `--help`


nomeata:
Hi,

the output would probably get rather large, but I still would like my users to enter `foo --help` and learn not just about global options and a concise list of subcommands, but also a full description of each command, including their options, so that they do not have to run `foo subcommand --help` for each of them.

Thanks,
Joachim

HuwCampbell:
This seems like a reasonable request, but it's not trivial.
Given how subparsers are implemented, the top parser knows nothing about it's children; and they of their parent. It's tightly bound to #142 in that respect. So it might require changing the way context is dragged up and down again.

chreekat:
fwiw, I'd like a middle ground, where `foo --help` looks like:

    sdb.hs: a wrapper to set up environment variables and run various
    commands for your local postgres database for Snowdrift.coop
    development.

    Usage:

        sdb.hs ACTION [ARGS]

    Where ACTION may be one of:

        init              initialize the database cluster and databases
        start             start the cluster
        stop              stop  the cluster
        clean             "rm -rf" the whole cluster
        reset             re-init the cluster and databases
        export            create data dumps with pg_dump

        # Expert commands:
        pg_ctl            run Postgres' pg_ctl(1) utility
        psql DATABASE     connect to DATABASE with psql

chreekat:
ok, that's more of a fork in the road than a middle ground, but ... :)

HuwCampbell:
So @chreekat your think should be possible after #197

HuwCampbell:
Addressed in #385 

Feel free to follow up if you think something isn't covered.

PR #385, opened 2020-05-15
Topic/global options

HuwCampbell:
This PR adds global options to the help text of sub commands.

This is not finished, but worth commenting on now.

Addresses #175, #294, and #359.


HuwCampbell:
I'm actually extremely pleased with how well this has turned out; and think that once this merged it could be worth a new release. I'm surprised I didn't think of it earlier.

The interesting edge cases I can think of:

- Completed options from the parent parser won't be shown.
  We use the context at the parse time of the subcommand. I don't think this is a big deal; people probably know the arguments they've already used.

- The options may not be accessible _right now_. For example, with  no-backtrack, one can't just add the option on the end as it's not in scope anymore. I don't think this is a big deal as they're clearly marked as a _global_ options.

Issue #379, opened 2020-04-16
cmd --help subcmd is not the same as cmd subcmd --help


shmish111:
It seems a bit wrong that `cmd --help subcmd` effectively ignores `subcmd`. I think it should either fail or provide the help for the subcmd (ideally the second)

HuwCampbell:
I'm inclined to agree.

merijn:
This issue was previously discussed here: #294

merijn:
And the former version definitely *cannot* be allowed to fail until #294 is resolved, or the user will have no way to query the help for `cmd`, since those aren't shown as part of `cmd subcmd --help` currently.

shmish111:
Yes, #294 is also pretty important to me and is related (although definitely not the same).

HuwCampbell:
I had a think and figured this one out on top of #385.

For the `cabal` test example, optparse now produces this:

```
*Examples.Cabal> :main --help install
Usage: <interactive> install [--enable-tests] [-f|--flags FLAGS] [--reinstall] 
                             [--force-reinstall]
  Installs a list of packages

Available options:
  --enable-tests           Enable compilation of test suites
  -f,--flags FLAGS         Enable the given flag
  -h,--help ARG            Show this help text

Global options:
  -v,--verbose LEVEL       Set verbosity to LEVEL
  --version                Print version information
```

merijn:
I've been playing with this over the weekend and the results look good here :)

As for the consistency issue of how `--help` displays and how you can specify flags, maybe the global options should (by default) be hidden, unless backtracking is enabled to avoid user confusion?

HuwCampbell:
> I've been playing with this over the weekend and the results look good here :)

That's great to hear.

Yes, I definitely considered changing the defaults. I'm happy to discuss, but there were quite a few changes just now. Please post code snippets with expected and desired so we can talk on the same page.

I'm going to close this issue because I think it's resolved, but feel free to post here.

HuwCampbell:
So just to summarise @shmish111.

Because people often just tab to the front or end or anywhere and bash `--help`, I need it to "do the right thing and show the help screen".

The behaviour now is to show a subcommand help screen if we can find one in the parent parser as we would have rendered its help screen. If we can't find one, we just render the help where we are.

shmish111:
Awesome work, optparse is such a fantastic project (surely there's not better CLI parsing library out there for any language) and its great to see how user feedback is acted upon in such a thoughtful and timely manner.

PR #386, opened 2020-05-17
Allow --help to take as an argument a command name.

HuwCampbell:
Allow --help to take as an argument a subcommand name. If the argument to --help is an actual subcommand, display the subcommand's help text instead.

Addressing #379 and #175 . 

CC: @shmish111 @merijn


HuwCampbell:
Sorry, this isn't quite there yet.

PR #399, opened 2020-06-18
Update changelog for release

HuwCampbell:


HuwCampbell:
I've been doing some testing with different tools I've found. I think the global options display should be opt-in rather than opt-out.

HuwCampbell:
I've uploaded a package candidate.

PR #408, opened 2020-10-14
Use bash from $SHELL when it makes sense

roberth:
This fixes file completion in cases where a virtual environment with a non-interactive bash is loaded with direnv, nix-shell or similar.

### DIFF
--- a/src/Options/Applicative/Builder/Completer.hs
+++ b/src/Options/Applicative/Builder/Completer.hs
@@ -9,8 +9,9 @@ module Options.Applicative.Builder.Completer
 import Control.Applicative
 import Prelude
 import Control.Exception (IOException, try)
-import Data.List (isPrefixOf)
+import Data.List (isPrefixOf, isSuffixOf)
 import System.Process (readProcess)
+import System.Environment (lookupEnv)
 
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Hi thanks for contributing to optparse!

I'll be honest, I'm very conflicted on this, though happy to discuss further.

What you've shown in your terminal printout is awful! We just regurgitate stderr of the base process into the user's shell when really, they don't want that. What I'm thinking is changing the `bashCompleter` function to something like this:

```haskell
bashCompleter :: String -> Completer
bashCompleter action = Completer $ \word -> do
  let cmd = unwords ["compgen", "-A", action, "--", requote word]
  (ec, _, result) <-
    readProcessWithExitCode "bash" ["-c", cmd] ""

  return $
    if ec == ExitSuccess then
      lines result
    else
      []
```

Here we get rid of a exception handler and won't spew error messages to their terminal when they hit tab. Your solution won't fix _that_ in the general case, but could reduce it for some.

I don't know, is that acceptable?

roberth:
A change like that will make it _practically_ impossible to troubleshoot, unless someone is dead set on solving a minor inconvenience.

PR #409, opened 2020-11-06
Guard process package with a flag

Bodigrim:
I'm trying to employ `optparse-applicative` for an in-package test-suite for `bytestring`. The main difficulty is to avoid circular dependencies, because `bytestring` is pretty ubiquitous. It seems that `process` package is used by `optparse-applicative` only for Bash autocompletion, but incurs a relatively large dependency graph, involving `bytestring` on several occasions.

This PR introduces a cabal flag, allowing to disable `process` when needed. 

Current dependency graph:
before

With `-f-process`:
after

### DIFF
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -60,6 +60,11 @@ source-repository head
   type:     git
   location: https://github.com/pcapriotti/optparse-applicative.git
 
+flag process
+  description:
+    Depend on the process package for Bash autocompletion
+  default: True
+
 library
   hs-source-dirs:      src
   ghc-options:         -Wall
@@ -92,8 +97,10 @@ library
   build-depends:       base                            == 4.*
                      , transformers                    >= 0.2 && < 0.6
                      , transformers-compat             >= 0.3 && < 0.7
-                     , process                         >= 1.0 && < 1.7
                      , ansi-wl-pprint                  >= 0.6.8 && < 0.7
+
+  if flag(process)
+    build-depends:     process                         >= 1.0 && < 1.7
 
   if !impl(ghc >= 8)
     build-depends:     semigroups                      >= 0.10 && < 0.20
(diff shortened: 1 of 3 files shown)


HuwCampbell:
Cool. As mentioned earlier I think this should be fine, there *might* be another package with a `process` flag, would that cause issues?

Bodigrim:
I don't think so, Cabal flags are specified and applied package-wise, not global. For example, one can invoke
```
cabal build --constraint 'optparse-applicative -process' --constraint 'foobar +process'
```

HuwCampbell:
With regards to the test suite including bytestring, I was already planning on removing it.

You can put that commit into this PR as well.

HuwCampbell:
The other question is do you need a release any time soon?

I have a change to tabulation in the usage text on master which would force a version bump. I'm a bit loath to do this more than once a year as optparse is pretty low on the dependency graph and it causes a lot of thrash.

So if it'll help, I'll rebase these commits over the 0.16 tag and put out a 0.16.1.

Regards,
Huw

Bodigrim:
> You can put that commit into this PR as well.

Done.

> So if it'll help, I'll rebase these commits over the 0.16 tag and put out a 0.16.1.

This would be perfect, thanks.

HuwCampbell:
Cool. I'll have a go on the weekend.

Bodigrim:
@HuwCampbell just a gentle reminder about this PR.

HuwCampbell:
I've rebased your commits onto the 0.16.0 tag and updated the change log. It's on a branch called update/0.16.1

HuwCampbell:
https://hackage.haskell.org/package/optparse-applicative-0.16.1.0/candidate

HuwCampbell:
It's up

Bodigrim:
Thanks a ton, hugely appreciated.

Issue #284, opened 2017-12-28
Parsers that consume multiple arguments


chris-martin:
Is there a way to write a parser for an option that takes multiple command-line arguments?

For example, if I wanted the command-line args

    -p one two -p three four

to be parsed into a result like

    [("one", "two"), ("three", "four")]

HuwCampbell:
Unfortunately not; I have some ideas for this could be done, but backwards compatibility is always an issue and I haven't sat down and had a real crack at it yet.

HuwCampbell:
It's also not really recommended by the [posix command line specifications](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).

You can always do something like what git does with key maps `[-c name=value]` for instance.

Maybe I can help design something else if you can share your problem?

chris-martin:
I'm making a simple tool for string replacements, where each pair ("abc", "xyz") means "where you see string 'xyz', replace it with 'abc';" like `sed s/abc/xyz/g`. Unfortunately the general-purpose nature of this tool makes the `name=value` strategy awkward, since the character `=` (or anything else we pick as the delimiter) could very well be part of the input, and I really don't want to force the user to escape anything beyond what their shell requires.

HuwCampbell:
I actually think that that may be a pretty good API call all things being equal.

Sorry for not getting back to you @chris-martin, I didn't have a good solution apart from just use many `-e` options with sed (like, why not?).

At the moment, we have
```haskell
newtype ReadM a = ReadM
  { unReadM :: ReaderT String (Except ParseError) a }
```

which this obviously doesn't support. The last time we changed this API however (see #108 in 2014), it caused quite a lot of pain, anger, and frustration, and even lead to a few libraries jumping ship.

As optparse is quite low on the pile of haskell dependencies, I'm pretty cautious about large changes to its types like this. Although I think it's a bit daft, I have seen people using the `ReaderT` constructor in their `ReadM` definitions; more pernicious, is that people might just use `str` multiple times, which would still compile but be wrong if I were to make this API change.

So I'm not against this, by all means, but I'm not going to rush it through.

Before going further, I would need to figure out the ramifications.

Cheers,
Huw

chris-martin:
> like, why not?

Because my program does this:

```
$ echo "banana" | text-replace --delimiter " " --mapping "a b b a"
abnbnb
```

not this:

```
[...]

Issue #271, opened 2017-06-25
Options with multiple arguments


wdanilo:
Hi!
Is it possible to parse options with multiple arguments using optparse?
Of course we are easily able to express it using applicative style, like:

```
multiArg = (,) <$> strOption   (hidden <> long "pass" <> metavar "switch name" <> help "Switch passes execution.")
               <*> strArgument (internal <> metavar "name")
```

which works fine, but as noted in the example code, we have to use `internal` modifier because this option is wrongly displayed in both help and usage without it. But it is not the proper solution for this problem, unfortunatelly. When using `internal` modifier, when the option is not provided we get an error `Missing:` which is not helpful at all. 

PS. The metavar I use above "switch name" is just hack to display metavar of first arg and second arg when the second is hidden by `internal` modifier.

HuwCampbell:
It's not a supported use can no, as it's not really conforming with the unix command line conventions.

I don't know your use case, but if it's a configuration map or something similar, you might want to try a parser with `[-c name=value]`; a comma or colon separated list is also a common. It's easy to write these with `eitherReader` and Parsec (or friends).

wdanilo:
I understand it is not "standatd" unix command syntax, however it is possible to easily create such construction using the Applicative interface. This library is not called "optparse-unix" and I thought that everything we can create using applicatives should be supported here. In fact the above formula is a very simple one and the generated help text is just inappropriate, so I still think we should consider this as a bug. I will however understand if you've got another opinion in this topic – then we can close this issue.

HuwCampbell:
The parser you've specified is indeed allowed and makes sense, but it's not an option which takes two arguments. Rather, it's a positional argument and an option. Options can go before or after the positional argument (with noIntersperse on) and be written in any order with other options. So a user can write
```
program helpValue --pass switchValue
```
or
```
program --pass switchValue --someOtherFlag helpValue
```
instead of
```
program --pass switchValue helpValue
```
and all are equally valid (which is the intentional behaviour of options and arguments in the library).

Does that make sense?

HuwCampbell:
Feel free to follow up

PR #415, opened 2021-03-05
Add biOption for two-argument options

roberth:
This is a potential solution to #271 and #284, observing that both issues only seem to want to parse a fixed number of two arguments per option. That certainly applies to my use case as well.

I don't think this solution is pretty, but it is backwards compatible and it does the job.

### DIFF
--- a/src/Options/Applicative.hs
+++ b/src/Options/Applicative.hs
@@ -65,6 +65,8 @@ module Options.Applicative (
   strOption,
   option,
 
+  biOption,
+
   strArgument,
   argument,
 
@@ -93,6 +95,7 @@ module Options.Applicative (
   showDefaultWith,
   showDefault,
   metavar,
+  metavar2,
   noArgError,
   hidden,
   internal,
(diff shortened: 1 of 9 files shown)


HuwCampbell:
Interesting, but yeah I agree it's a bit janky.

So I'm not sure (using GADTs is almost certainly a no though as I'm trying to keep optparse very lean).

I think if people like yourself are really needing this, the best course of action is to resurrect #284, but make the behaviour opt-in under a new `Mod OptionFields a` modifier which turns on the variable length consumption.

That way people can't easily foot-gun themselves, and no-one's code will break.

MagicRB:
@roberth did you not end up needing this, trying to translate a bash script into Haskell and need to parse `--override-input nixpkgs something`, you'll never guess what package manager this is for.

MagicRB:
The most temporary solutions become the most permenant, or something. I'll use your fork then, because `--override-input nixpkgs=foobar` is everything but nice

PR #416, opened 2021-03-06
Fix Semigroup instance of ParserHelp

jhrcek:


src/Options/Applicative/Help/Types.hs
```diff
@@ -28,10 +28,10 @@ instance Monoid ParserHelp where
   mappend = (<>)
 
 instance Semigroup ParserHelp where
-  (ParserHelp e1 s1 h1 u1 d1 b1 g1 f1) <> (ParserHelp e2 s2 h2 d2 u2 b2 g2 f2)
+  (ParserHelp e1 s1 h1 u1 d1 b1 g1 f1) <> (ParserHelp e2 s2 h2 u2 d2 b2 g2 f2)
```
jhrcek, line 31: d2 and u2 (standing for Description and Usage) of the 2nd Parser help were swapped.
Base on ParserDefinition Usage goes first, then description.

HuwCampbell:
Thanks.

PR #417, opened 2021-03-15
Add helperWith to customize the helper

roberth:
Allows some customization of the helper option.

I use this to make the option internal, removing it from the help text and completions. Doing so makes completion more effective because it gets rid of the only short option. Short options have no place in a modern command in my opinion, because a single letter carries next to no meaning, which is error prone and bad for script readability. In the interactive case, tab completion still saves the user the extra typing.

### DIFF
--- a/src/Options/Applicative/Extra.hs
+++ b/src/Options/Applicative/Extra.hs
@@ -4,6 +4,7 @@ module Options.Applicative.Extra (
   --
   -- | This module contains high-level functions to run parsers.
   helper,
+  helperWith,
   hsubparser,
   execParser,
   customExecParser,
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Looks good.

I would:
- not export in the top level module; and
- move `hidden` back to helperWith.

The `hidden` modifier can be overridden by `internal`; so I think it's worthwhile keeping the default where it is.

roberth:
> * not export in the top level module; and
> * move `hidden` back to helperWith.

Done.

> The `hidden` modifier can be overridden by `internal`; so I think it's worthwhile keeping the default where it is.

I don't think `hidden` can be undone, can it? I'm ok with that though.

HuwCampbell:
It can be actually. Builders are applied in an order, so a builder which does this:

```haskell
visible :: Mod f a
visible = optionMod $ \p -> p { propVisibility = Visible }
```

will do the trick.

HuwCampbell:
Ta

PR #418, opened 2021-03-29
fix noncompilable readme example

### LINKED ISSUE #286: Docs about commands are slightly confusing
In https://github.com/pcapriotti/optparse-applicative/blob/ff5889543792eac2f633d1bb0c2176087f1e5f68/README.md#commands it says:

---

```haskell
subparser
  ( command "add" (info addOptions ( progDesc "Add a file to the repository" ))
 <> command "commit" (info commitOptions ( progDesc "Record changes to the repository" ))
  )
```

[...]

> Note that all the parsers appearing in a command need to have the same type. For this reason, it is often best to use a sum type which has the same structure as the command itself. For example, for the parser above, you would define a type like:

```haskell
data Options = Options
  { optCommand :: Command
  , ... }

data Command
  = Add AddOptions
  | Commit CommitOptions
  ...
```

---

However, this is confusing, because from the writing, one would assume that `addOptions :: AddOptions` and `commitOptions :: CommitOptions`.

But if that is the case, the example would not compile, as the two `command`s do NOT have the same type. A compiling example would be:

```haskell
subparser
  ( command "add" (info (Add <$> addOptions) ( progDesc "Add a file to the repository" ))
 <> command "commit" (info (Commit <$> commitOptions) ( progDesc "Record changes to the repository" ))
  )
```

I recommend to change the docs to show only this compiling example. To do so, I'd probably swap the order of the code listings, so that the `data`s are shown first and then how to parse them with `subparser`.

ibizaman:
fixes #286

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,8 +550,8 @@ modifier. For example,
 
 ```haskell
 subparser
-  ( command "add" (info (Add <$> addOptions) ( progDesc "Add a file to the repository" ))
- <> command "commit" (info (Commit <$> commitOptions) ( progDesc "Record changes to the repository" ))
+  ( command "add" (info addCommand ( progDesc "Add a file to the repository" ))
+ <> command "commit" (info commitCommand ( progDesc "Record changes to the repository" ))
   )
 ```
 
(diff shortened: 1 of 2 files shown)


pcapriotti:
It might be slightly nicer to replace `Add <$> addOptions` with simply `addCommand`, and similarly for `Commit`.

pcapriotti:
Merged. Thanks for the patch.

Issue #315, opened 2018-09-06
Powershell autompletion script


ShrykeWindgrace:
I use autocompletion scripts provided by `optparse-applicative` quite a lot in bash and zsh. Are there any plans to provide such scripts for powershell? Is is possible at all?

HuwCampbell:
I'm glad to hear that you're enjoying the bash and zsh ones. I don't have a Windows installation right now, so can't really look into it for you at the moment. I'll revisit if that changes, but I'm also open to PRs.

HuwCampbell:
Still don't have a windows machine. If you want to put up a PR though I can have a look. I think the biggest challenge is that we support bash functions for completions out of the box, and these will fail if bash isn't installed. User miles may vary in other words...

ShrykeWindgrace:
Hi! To be honest, I am not really familiar with inner workings of powershell scripts, I guess I'll manage something partially working soon, but I do not promise anything definitive.

ShrykeWindgrace:
A bit of necroposting: in fact, one does not need a windows machine to use powershell (per https://github.com/PowerShell/PowerShell, it's been cross-platform for quite some time).

A more useful remark for a user who has more time and desire to further this subject: it seems that `Register-ArgumentCompleter` with a carefully written `ScriptBlock` dynamically querying the desired executable with `--bash-completion-index` and `--bash-completion-word` behaves somewhat equivalent to current `--bash-completion-script`.

Again, I promise nothing, maybe in a year ( I know, right =))  I'll have a working solution.

ShrykeWindgrace:
And  a quick addendum:
```powershell
$scriptblock = {
  param($commandName, $parameterName, $stringMatch)
    $result = & invoke-Expression "myExecutable.exe --bash-completion-index 0 --bash-completion-word $commandName"
    $result
    }
Register-ArgumentCompleter -Native -CommandName myExecutable.exe -ScriptBlock $scriptBlock
```
This script already gives a rudimentary tab-completion for names of `long` options - it only handles the cases `--part-of-option`, it does not work for `-`, `--`, nor for `short` options, neither it can validate the possible values of a parameter.

PR #420, opened 2021-04-16
Add powershell autocompletion script generation

ShrykeWindgrace:
This almost completely addresses issue #315. I leverage the same mechanism as other completion scripts. There are some quirks however:

- on a windows machine, due to limitations of [getProgName](https://www.stackage.org/haddock/lts-17.9/base-4.14.1.0/System-Environment.html#v:getProgName) the script will work for `foo.exe`, not `foo`. I mention  this in updated Readme. I will investigate the problem further, maybe it is safe to strip trailling `.exe`.
- powershell-5 has some limitations - it does not trigger autocompletion on `foo --`. This is a known issue, and it will not be fixed. Powershell-7 has no such limitations.

A review of my code and readme wording is most welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -791,10 +791,10 @@ French commands:
   au-revoir                Say goodbye
 ```
 
-## Bash, Zsh, and Fish Completions
+## Bash, Zsh, Fish, and Pwsh Completions
 
 `optparse-applicative` has built-in support for the completion of
-command line options and arguments in bash, zsh, and fish shells.
+command line options and arguments in bash, zsh, fish, and pwsh shells.
 Any parser, when run using the `execParser` family of functions,
 is automatically extended with a few (hidden) options for the
 completion system:
@@ -817,13 +817,25 @@ completion system:
 
  - `--fish-completion-script`: which is analogous for fish shell;
 
+ - `--pwsh-completion-script`: which is analogous for powershell/pwsh.
+   You might want to generate a script and then dot-source it as
+
+   ```console
+   PS> foo --pwsh-completion-script (Get-Command foo).Source >> _foo.ps1
+   PS> . _foo.ps1
+   ```
+   Note for windows users - this will generate completion script for `foo.exe`,
+   which is not equivalent to `foo` from tab-completion engine's point of view,
+   even though from usage point view they are equivalent. You might want to edit
+   the file `_foo.ps1` if you prefer to use `foo` and not `foo.exe`. 
+
  - `--bash-completion-index`, `--bash-completion-word`: internal options used
    by the completion script to obtain a list of possible completions for a
    given command line;
 
  - `--bash-completion-enriched`: a flag to tell the completion system to emit
    descriptions along with possible completions. This is used to provide help
-   along with the completion for `zsh` and `fish`.
+   along with the completion for `zsh`, `fish`, and `pwsh`.
 
 ### Actions and completers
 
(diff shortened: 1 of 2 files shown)


HuwCampbell:
Thanks for the contribution. I'll test this out soon.

HuwCampbell:
I gave it a go on a unix system. Thanks again for putting in the time.

It looks like it's functioning _ok_, but I found a few quirks. I'm not on windows though, so your experience might be different.

Tab completions would always start with `./` if I don't write anything. So if there's a list of subcommands available, I need to type in a first letter for one of them? The default completer looks like it's overriding too much.

Options don't show their hyphens at the start. It looks like you've done this intentionally with a trim, but it strikes me as odd.

Option completion doesn't give a trailing space if the word is actually complete.

I can't seem to get it to show the tooltip. Again, maybe this is just the unix version.

ShrykeWindgrace:
> I gave it a go on a unix system. Thanks again for putting in the time.

Thanks for testing my code!

> 
> It looks like it's functioning _ok_, but I found a few quirks. I'm not on windows though, so your experience might be different.
> 
> Tab completions would always start with `./` if I don't write anything. So if there's a list of subcommands available, I need to type in a first letter for one of them? The default completer looks like it's overriding too much.

Do you mean that 
```console
$ foo [TAB]
```
results in `$ foo ./`? Or in `$ foo ./nameOfAFileInThisFolder`?

> 
> Options don't show their hyphens at the start. It looks like you've done this intentionally with a trim, but it strikes me as odd.

Indeed, I did it on purpose; though I do not have a really strong opinion on this, just a matter of taste. I can remove the trimming.

> 
> Option completion doesn't give a trailing space if the word is actually complete.

What's the scenario here?
```console
$ foo --full-option[TAB]
```
should append a space, if I understand correctly?

> 
> I can't seem to get it to show the tooltip. Again, maybe this is just the unix version.

This most probably comes from the default settings  of the `PSReadLine` module. This module is shipped by default for all pltaforms. Could you try putting
`Set-PSReadlineKeyHandler -Key Tab -Function MenuComplete` in your session or in your `$PROFILE`? I honestly don't understand why this setting is not the default one. If tooltips come back, I'll amend the Readme. This setting might also fix some other problems that you encountered. In the meanwhile, I'll test subcommands in a linux env.

ShrykeWindgrace:
Indeed, I observe the same problem with subcommands, I'll look into it.

Regarding the space after trying to complete an already complete word, we can fix that by artificially appending a space to the completion result.

andreasabel:
@HuwCampbell : Looks like @ShrykeWindgrace has pushed an update and waits for a re-review, or maybe I am misinterpreting the log here?  (Just passing by...)

ShrykeWindgrace:
@andreasabel Indeed, thanks for bumping the thread, I myself forgot about this PR =)

andreasabel:
For me, getting auto-completion for free was the killer argument to adopt `optparse-applicative`, so I think it is important that this feature is extended to a wider setting (meaning more shells, such as `pwsh`).

HuwCampbell:
Yep, sorry I haven't been too on the spot for this one.

The completion scripts have a tendancy of being really subtle. I remember when I was improving the completion scripts for 0.14 I was refining and manually testing for days (it was for fish, zsh, and bash though).

I'll take another look.

ShrykeWindgrace:
@HuwCampbell Hi! It's been quite a while since the inception of this PR and since the last interaction in this thread.
I understand perfectly well that during these stressful times carving a chunk of time to review this PR can be rather difficult.

So if it is indeed the case, could you give a rough estimation whether you'll have time for this in the foreseeable future? This PR is not a blocking feature for me, but I would welcome some new info.

Cheers!

ShrykeWindgrace:
Friendly ping =)

HuwCampbell:
Hi.

I've just rebased this. The only issue major I can see at the moment is that when completing files, if you go to a directory it does it as the whole work without a trailing slash, so you have to go back, add the slash and tab again from there instead of it adding it automatically.

PR #422, opened 2021-04-21
Fix warnings on recent GHC

pcapriotti:
Not sure if it's worth going through the trouble, but it might be nice to have no warnings on all supported GHC versions.


HuwCampbell:
LGTM.

PR #426, opened 2021-07-15
Convert to use prettyprinter library

newhoggy:
I created this PR to see what converting `optparse-applicative` to use `prettyprinter` library instead of `'ansi-wl-pprint` might look like.


HuwCampbell:
There are quite a lot or PRs here for pretty printing changes, it's a bit much.

This one changes all of the type signatures, which is not a reasonable burden for users. I'm going to close this one and #429 because we can't have a meaninful conversation spread all over the shop.

Issue #273, opened 2017-07-03
Consider using `prettyprinter` instead of `ansi-wl-pprint`


quchen:
Hi there,

tl;dr: I was wondering whether you would be open to adapting the `prettyprinter` for `optparse-applicative` (instead of `ansi-wl-pprint`).

---

# Longer, want to read

I recently released a vastly overhauled version of our beloved Wadler/Leijen prettyprinter, called [`prettyprinter`](https://hackage.haskell.org/package/prettyprinter).

I believe many people were unhappy with certain aspects of `wl-pprint`, and many of them felt like adding »that one thing«, which left us with a plethora of Wadler/Leijen prettyprinters, each one with one shortcoming less, but none of them lacking most of the other features. `prettyprinter` was designed to remedy this, by combining all the popular features (as demanded by multiple popular projects, including e.g. GHC, Idris, Trifecta, and yes, optparse-applicative), and adding excellent documentation and migration advice ontop of it to make the process as simple as possible.

Feedback to `prettyprinter` has been very positive on release, and it is even under consideration for use internally in GHC.

# What’s to be gained?

Why I’m asking here? Well, `optparse-applicative` is a very popular library, and I see it as one of the roots of `ansi-wl-pprint` usage in the Haskell ecosystem. I think there is much to be gained (in terms of pain avoided) by having a standard library for prettyprinting that does all the things and does it well; users should not have to have to use an ANSI terminal dependency without the capability of annotations when they also want good command line parameter handling.

Having `optparse-applicative` switch to using `prettyprinter` would be a huge milestone to unification of all the prettyprinters, ideally deprecating the others in its favor, should it be able to gain enough traction.

# What’s to be lost?

There are conversion functions from `Doc` (`ansi-wl-pprint`) to `Doc` (`prettyprinter`) and back, so users that *really* want/have to work with `ansi-wl-pprint` can still use it. Apart from this, I don’t expect any problems worth mentioning.

The conversion functions are found in the not yet released `prettyprinter-convert-ansi-wl-pprint` package, which will be released as soon as `ansi-wl-pprint` makes its next Hackage release (which I’m working on these days).

# Why this prettyprinter, and not one of the others?

This is what `prettyprinter`’s readme is all about: [README.md](https://github.com/quchen/prettyprinter/blob/5a87842910c2031f1e95426932caf93726065102/README.md)

[...]

PR #428, opened 2021-07-21
Pretty printer library switch with ANSI compatibility

newhoggy:
This PR modifiers the pretty printer to improve the structure of the output for readability.

Included in the change is the switch from `ansi-wl-pprint` to `prettyprinter`.  The switch allows for the use of annotations, which allows for tracing the structure of the document for ease of debugging.

I'm looking for interest in merging improved formatting and switching to `prettyprinter`.

### DIFF
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -88,19 +88,22 @@ library
                      , Options.Applicative.Common
                      , Options.Applicative.Extra
                      , Options.Applicative.Help
+                     , Options.Applicative.Help.Ann
                      , Options.Applicative.Help.Chunk
                      , Options.Applicative.Help.Core
                      , Options.Applicative.Help.Levenshtein
                      , Options.Applicative.Help.Pretty
+                     , Options.Applicative.Help.Style
                      , Options.Applicative.Help.Types
                      , Options.Applicative.NonEmpty
                      , Options.Applicative.Types
                      , Options.Applicative.Internal
 
   build-depends:       base                            == 4.*
+                     , ansi-terminal                   >= 0.4.0
+                     , prettyprinter                   >= 1.7.1 && < 1.8
                      , transformers                    >= 0.2 && < 0.7
                      , transformers-compat             >= 0.3 && < 0.8
-                     , ansi-wl-pprint                  >= 0.6.8 && < 0.7
 
   if flag(process)
     build-depends:     process                         >= 1.0 && < 1.7
--- a/src/Options/Applicative/Help/Core.hs
+++ b/src/Options/Applicative/Help/Core.hs
@@ -35,8 +35,8 @@ import Prelude hiding (any)
 
 import Options.Applicative.Common
 import Options.Applicative.Types
-import Options.Applicative.Help.Pretty
 import Options.Applicative.Help.Chunk
+import Options.Applicative.Help.Pretty
 
 -- | Style for rendering an option.
 data OptDescStyle
@@ -157,6 +157,7 @@ foldTree prefs s (MultNode xs) =
   where
     mult_wrap [_] = NeverRequired
     mult_wrap _ = MaybeRequired
+
 foldTree prefs s (AltNode b xs) =
   (\x -> (x, NeverRequired))
     . fmap groupOrNestLine
--- a/src/Options/Applicative/Help/Pretty.hs
+++ b/src/Options/Applicative/Help/Pretty.hs
@@ -1,10 +1,70 @@
 {-# LANGUAGE CPP #-}
+
 module Options.Applicative.Help.Pretty
-  ( module Text.PrettyPrint.ANSI.Leijen
+  ( module PP
+  , Ann(..)
+  , Doc
   , (.$.)
   , groupOrNestLine
   , altSep
   , hangAtIfOver
+
+  , enclose
+  , parens
+  , brackets
+  , hang
+  , indent
+  , nest
+
+  , text
+  , plain
+  , deunderline
+  , underline
+  , debold
+  , bold
+  , ondullwhite
+  , onwhite
+  , ondullcyan
+  , oncyan
+  , ondullmagenta
+  , onmagenta
+  , ondullblue
+  , onblue
+  , ondullyellow
+  , onyellow
+  , ondullgreen
+  , ongreen
+  , ondullred
+  , onred
+  , ondullblack
+  , onblack
+  , dullwhite
+  , white
+  , dullcyan
+  , cyan
+  , dullmagenta
+  , magenta
+  , dullblue
+  , blue
+  , dullyellow
+  , yellow
+  , dullgreen
+  , green
+  , dullred
+  , red
+  , dullblack
+  , black
+
+  -- TODO Remove these
+  -- , (<$>)
+  , (</>)
+  , (<$$>)
+  , (<//>)
+  , string
+
+  , isEffectivelyEmpty
+
+  , renderShowS
   ) where
 
 import           Control.Applicative
@@ -12,15 +72,19 @@ import           Control.Applicative
 import           Data.Semigroup ((<>))
 #endif
 
-import           Text.PrettyPrint.ANSI.Leijen hiding ((<$>), (<>), columns)
-import           Text.PrettyPrint.ANSI.Leijen.Internal (Doc (..), flatten)
-import qualified Text.PrettyPrint.ANSI.Leijen as PP
+import           Options.Applicative.Help.Ann
+import           Prettyprinter hiding ((<>), Doc, enclose, parens, brackets, hang, indent, nest)
+import qualified Prettyprinter as PP
+import qualified Prettyprinter.Internal as PPI
+import           Prettyprinter.Render.String (renderShowS)
+import qualified Options.Applicative.Help.Style as S
 
 import           Prelude
 
-(.$.) :: Doc -> Doc -> Doc
-(.$.) = (PP.<$>)
+type Doc = PPI.Doc Ann
 
+(.$.) :: Doc -> Doc -> Doc
+(.$.) x y = x <> line <> y
 
 -- | Apply the function if we're not at the
 --   start of our nesting level.
@@ -38,8 +102,8 @@ ifAtRoot =
 --   start of our nesting level.
 ifElseAtRoot :: (Doc -> Doc) -> (Doc -> Doc) -> Doc -> Doc
 ifElseAtRoot f g doc =
-  Nesting $ \i ->
-    Column $ \j ->
+  PPI.Nesting $ \i ->
+    PPI.Column $ \j ->
       if i == j
         then f doc
         else g doc
@@ -52,10 +116,25 @@ ifElseAtRoot f g doc =
 --   group.
 groupOrNestLine :: Doc -> Doc
 groupOrNestLine =
-  Union
+  PPI.Union
     <$> flatten
     <*> ifNotAtRoot (line <>) . nest 2
+  where flatten :: Doc -> Doc
+        flatten doc = case doc of
+          PPI.FlatAlt _ y     -> flatten y
+          PPI.Cat x y         -> PPI.Cat (flatten x) (flatten y)
+          PPI.Nest i x        -> PPI.Nest i (flatten x)
+          PPI.Line            -> PPI.Fail
+          PPI.Union x _       -> flatten x
+          PPI.Column f        -> PPI.Column (flatten . f)
+          PPI.WithPageWidth f -> PPI.WithPageWidth (flatten . f)
+          PPI.Nesting f       -> PPI.Nesting (flatten . f)
+          PPI.Annotated ann x -> PPI.Annotated ann (flatten x)
 
+          x@PPI.Fail   -> x
+          x@PPI.Empty  -> x
+          x@PPI.Char{} -> x
+          x@PPI.Text{} -> x
 
 -- | Separate items in an alternative with a pipe.
 --
@@ -69,8 +148,7 @@ groupOrNestLine =
 --   next line.
 altSep :: Doc -> Doc -> Doc
 altSep x y =
-  group (x <+> char '|' <> line) <//> y
-
+  group (x <+> pretty "|" <> line) <> softline' <> y
 
 -- | Printer hacks to get nice indentation for long commands
 --   and subcommands.
@@ -85,8 +163,186 @@ altSep x y =
 --   the starting column, and it won't be indented more.
 hangAtIfOver :: Int -> Int -> Doc -> Doc
 hangAtIfOver i j d =
-  Column $ \k ->
+  PPI.Column $ \k ->
     if k <= j then
       align d
     else
       linebreak <> ifAtRoot (indent i) d
+
+(</>) :: Doc -> Doc -> Doc
+(</>) x y = x <> softline <> y
+
+(<$$>) :: Doc -> Doc -> Doc
+(<$$>) x y = x <> linebreak <> y
+
+(<//>) :: Doc -> Doc -> Doc
+(<//>) x y = x <> softbreak <> y
+
+linebreak :: Doc
+linebreak = flatAlt line mempty
+
+softbreak :: Doc
+softbreak = group linebreak
+
+-- | Traced version of 'PP.string'.
+string :: String -> Doc
+string = PP.pretty
+
+-- | Traced version of 'PP.parens'.
+parens :: Doc -> Doc
+parens = PP.parens
+
+-- | Traced version of 'PP.brackets'.
+brackets :: Doc -> Doc
+brackets = PP.brackets
+
+-- | Traced version of 'PP.enclose'.
+enclose
+    :: Doc -- ^ L
+    -> Doc -- ^ R
+    -> Doc -- ^ x
+    -> Doc -- ^ LxR
+enclose = PP.enclose
+
+-- | Traced version of 'PP.hang'.
+hang :: Int -> Doc -> Doc
+hang = PP.hang
+
+-- | Traced version of 'PP.nest'.
+nest :: Int -> Doc -> Doc
+nest = PP.nest
+
+-- | Traced version of 'PP.indent'.
+indent :: Int -> Doc -> Doc
+indent = PP.indent
+
+-- | Determine if the document is empty when rendered
+isEffectivelyEmpty :: Doc -> Bool
+isEffectivelyEmpty doc = case doc of
+  PPI.Fail -> True
+  PPI.Empty -> True
+  PPI.Char _ -> False
+  PPI.Text _ _ -> False
+  PPI.Line -> False
+  PPI.FlatAlt _ d -> isEffectivelyEmpty d
+  PPI.Cat a b -> isEffectivelyEmpty a && isEffectivelyEmpty b
+  PPI.Nest _ d -> isEffectivelyEmpty d
+  PPI.Union _ d -> isEffectivelyEmpty d
+  PPI.Column _ -> True
+  PPI.WithPageWidth _ -> False
+  PPI.Nesting _ -> False
+  PPI.Annotated _ d -> isEffectivelyEmpty d
+
+text :: String -> Doc
+text = pretty
+
+plain :: Doc -> Doc
+plain = id
+
+deunderline :: Doc -> Doc
+deunderline = id
+
+underline :: Doc -> Doc
+underline = annotate (AnnStyle S.underlined)
+
+debold :: Doc -> Doc
+debold = id
+
+bold :: Doc -> Doc
+bold = annotate (AnnStyle S.bold)
+
+ondullwhite :: Doc -> Doc
+ondullwhite = annotate (AnnStyle (S.bgColorDull S.White))
+
+onwhite :: Doc -> Doc
+onwhite = annotate (AnnStyle (S.bgColor S.White))
+
+ondullcyan :: Doc -> Doc
+ondullcyan = annotate (AnnStyle (S.bgColorDull S.Cyan))
+
+oncyan :: Doc -> Doc
+oncyan = annotate (AnnStyle (S.bgColor S.Cyan))
+
+ondullmagenta :: Doc -> Doc
+ondullmagenta = annotate (AnnStyle (S.bgColorDull S.Magenta))
+
+onmagenta :: Doc -> Doc
+onmagenta = annotate (AnnStyle (S.bgColor S.Magenta))
+
+ondullblue :: Doc -> Doc
+ondullblue = annotate (AnnStyle (S.bgColorDull S.Blue))
+
+onblue :: Doc -> Doc
+onblue = annotate (AnnStyle (S.bgColor S.Blue))
+
+ondullyellow :: Doc -> Doc
+ondullyellow = annotate (AnnStyle (S.bgColorDull S.Yellow))
+
+onyellow :: Doc -> Doc
+onyellow = annotate (AnnStyle (S.bgColor S.Yellow))
+
+ondullgreen :: Doc -> Doc
+ondullgreen = annotate (AnnStyle (S.bgColorDull S.Green))
+
+ongreen :: Doc -> Doc
+ongreen = annotate (AnnStyle (S.bgColor S.Green))
+
+ondullred :: Doc -> Doc
+ondullred = annotate (AnnStyle (S.bgColorDull S.Red))
+
+onred :: Doc -> Doc
+onred = annotate (AnnStyle (S.bgColor S.Red))
+
+ondullblack :: Doc -> Doc
+ondullblack = annotate (AnnStyle (S.bgColorDull S.Black))
+
+onblack :: Doc -> Doc
+onblack = annotate (AnnStyle (S.bgColor S.Black))
+
+dullwhite :: Doc -> Doc
+dullwhite = annotate (AnnStyle (S.colorDull S.White))
+
+white :: Doc -> Doc
+white = annotate (AnnStyle (S.color S.White))
+
+dullcyan :: Doc -> Doc
+dullcyan = annotate (AnnStyle (S.colorDull S.Cyan))
+
+cyan :: Doc -> Doc
+cyan = annotate (AnnStyle (S.color S.Cyan))
+
+dullmagenta :: Doc -> Doc
+dullmagenta = annotate (AnnStyle (S.colorDull S.Magenta))
+
+magenta :: Doc -> Doc
+magenta = annotate (AnnStyle (S.color S.Magenta))
+
+dullblue :: Doc -> Doc
+dullblue = annotate (AnnStyle (S.colorDull S.Blue))
+
+blue :: Doc -> Doc
+blue = annotate (AnnStyle (S.color S.Blue))
+
+dullyellow :: Doc -> Doc
+dullyellow = annotate (AnnStyle (S.colorDull S.Yellow))
+
+yellow :: Doc -> Doc
+yellow = annotate (AnnStyle (S.color S.Yellow))
+
+dullgreen :: Doc -> Doc
+dullgreen = annotate (AnnStyle (S.colorDull S.Green))
+
+green :: Doc -> Doc
+green = annotate (AnnStyle (S.color S.Green))
+
+dullred :: Doc -> Doc
+dullred = annotate (AnnStyle (S.colorDull S.Red))
+
+red :: Doc -> Doc
+red = annotate (AnnStyle (S.color S.Red))
+
+dullblack :: Doc -> Doc
+dullblack = annotate (AnnStyle (S.colorDull S.Black))
+
+black :: Doc -> Doc
+black = annotate (AnnStyle (S.color S.Black))
--- a/src/Options/Applicative/Help/Style.hs
+++ b/src/Options/Applicative/Help/Style.hs
@@ -1,0 +1,126 @@
+{-# LANGUAGE CPP #-}
+
+module Options.Applicative.Help.Style
+  ( SetStyle (..)
+  , ColorIntensity (..)
+  , Layer (..)
+  , ConsoleIntensity (..)
+  , Underlining (..)
+  , Italicized (..)
+  , Color (..)
+  , color
+  , bgColor
+  , colorDull
+  , bgColorDull
+  , bold
+  , underlined
+  , italicized
+  , styleToRawText
+  , defaultStyle
+  ) where
+
+import Control.Applicative
+import Data.Maybe
+#if __GLASGOW_HASKELL__ <= 802
+import Data.Semigroup hiding (option)
+#endif
+import System.Console.ANSI (ConsoleIntensity (..), ColorIntensity (..), Underlining (..))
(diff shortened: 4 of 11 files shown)


newhoggy:
Before:

```
Usage: cardano-cli transaction build-raw [--byron-era | --shelley-era |
                                           --allegra-era | --mary-era |
                                           --alonzo-era] (--tx-in TX-IN
                                         [--tx-in-script-file FILE
                                           [
                                             (--tx-in-datum-file FILE |
                                               --tx-in-datum-value JSON VALUE)
                                             (--tx-in-redeemer-file FILE |
                                               --tx-in-redeemer-value JSON VALUE)
                                             --tx-in-execution-units (INT, INT)]])
                                         [--tx-in-collateral TX-IN]
                                         [--tx-out ADDRESS VALUE
                                           [--tx-out-datum-hash HASH]]
                                         [--mint VALUE (--mint-script-file FILE
                                           [
                                             (--mint-redeemer-file FILE |
                                               --mint-redeemer-value JSON VALUE)
                                             --mint-execution-units (INT, INT)])]
                                         [--invalid-before SLOT]
                                         [--invalid-hereafter SLOT]
                                         [--fee LOVELACE]
                                         [--certificate-file CERTIFICATEFILE
                                           [--certificate-script-file FILE
                                             [
                                               (--certificate-redeemer-file FILE |

                                                 --certificate-redeemer-value JSON VALUE)
                                               --certificate-execution-units (INT, INT)]]]
                                         [--withdrawal WITHDRAWAL
                                           [--withdrawal-script-file FILE
                                             [
                                               (--withdrawal-redeemer-file FILE |

                                                 --withdrawal-redeemer-value JSON VALUE)
                                               --withdrawal-execution-units (INT, INT)]]]
                                         [--json-metadata-no-schema |
                                           --json-metadata-detailed-schema]
                                         [--auxiliary-script-file FILE]
                                         [--metadata-json-file FILE |
                                           --metadata-cbor-file FILE]
                                         [--genesis FILE |
                                           --protocol-params-file FILE]
                                         [--update-proposal-file FILE]
                                         --out-file FILE
```

After:

```
Usage: cardano-cli transaction build-raw  
            [ --byron-era
            | --shelley-era
            | --allegra-era
            | --mary-era
            | --alonzo-era
            ]
            ( --tx-in TX-IN
              [ --tx-in-script-file FILE
                [ (--tx-in-datum-file FILE | --tx-in-datum-value JSON VALUE)
                  ( --tx-in-redeemer-file FILE
                  | --tx-in-redeemer-value JSON VALUE
                  )
                  --tx-in-execution-units (INT, INT)
                ]
              ]
            )
            [--tx-in-collateral TX-IN]
            [--tx-out ADDRESS VALUE [--tx-out-datum-hash HASH]]
            [ --mint VALUE
              ( --mint-script-file FILE
                [ (--mint-redeemer-file FILE | --mint-redeemer-value JSON VALUE)
                  --mint-execution-units (INT, INT)
                ]
              )
            ]
            [--invalid-before SLOT]
            [--invalid-hereafter SLOT]
            [--fee LOVELACE]
            [ --certificate-file CERTIFICATEFILE
              [ --certificate-script-file FILE
                [ ( --certificate-redeemer-file FILE
                  | --certificate-redeemer-value JSON VALUE
                  )
                  --certificate-execution-units (INT, INT)
                ]
              ]
            ]
            [ --withdrawal WITHDRAWAL
              [ --withdrawal-script-file FILE
                [ ( --withdrawal-redeemer-file FILE
                  | --withdrawal-redeemer-value JSON VALUE
                  )
                  --withdrawal-execution-units (INT, INT)
                ]
              ]
            ]
            [--json-metadata-no-schema | --json-metadata-detailed-schema]
            [--auxiliary-script-file FILE]
            [--metadata-json-file FILE | --metadata-cbor-file FILE]
            [--genesis FILE | --protocol-params-file FILE]
            [--update-proposal-file FILE]
            --out-file FILE
```

newhoggy:
Annotations allow me to output HTML to describe the structure of the document that is being rendered to better understand layout bugs:

Screen Shot 2021-07-21 at 1 05 37 pm

newhoggy:
The idea behind the new style is it prefers compact display such as this:

```
[--json-metadata-no-schema | --json-metadata-detailed-schema]
```

But that if the length of the group should exceed the remaining space on the line, it should use structured multiline indented format:

```
[ ( --certificate-redeemer-file FILE
  | --certificate-redeemer-value JSON VALUE
  )
  --certificate-execution-units (INT, INT)
]
```

Or combinations where it makes sense:

```
[ (--mint-redeemer-file FILE | --mint-redeemer-value JSON VALUE)
  --mint-execution-units (INT, INT)
]
```

newhoggy:
I created another version of this PR without the tracing annotations so it's clearer where the substantive differences are.

https://github.com/pcapriotti/optparse-applicative/pull/429

newhoggy:
Mentioning https://github.com/pcapriotti/optparse-applicative/issues/273 and @quchen because the PR involves switching to `prettyprinter`.

Bodigrim:
Switching a pretty-printing library means that every downstream package has to adapt, because `Doc` is prominently used in `optparse-applicative` API. To be honest, incurring such toll on 800+ packages just to ease your debugging process does not seem reasonable to me. 

My understanding is that you have debugged the issue at hand. Is it possible to revert back to `ansi-wl-pprint`?

(I’m not a maintainer)

newhoggy:
I had a go yesterday, but it seems `ansi-wl-pprint` has different layout rules for its combinators and I wasn't successful.

I pointed out tracing mainly to show that annotations are actually really convenient and could be used to annotate documents for richer output.

Bodigrim:
> I pointed out tracing mainly to show that annotations are actually really convenient and could be used to annotate documents for richer output.

What specifically do annotations buy us in case of `optparse-applicative` in production setting? What kind of richer console output are we looking for?

> I had a go at [haskell/criterion#241](https://github.com/haskell/criterion/pull/241) to show that in this case the necessary changes aren't too bad.

Yes, indeed, it is usually possible to migrate between pretty-printing libraries. But nevertheless it means that downstream dependencies must make necessary amends and releases, and pre-existing packages must be revised with `optparse-applicative < 0.17`. This is quite a sizeable amount of work for everyone.

> To ease compatibility issues, we could minimise breaking changes. `optparse-applicative` could provide all the exports necessary to deal with documents.

I think that more re-exports of this sort could pose even bigger compatibility issues in a long term. Imagine that some combinator is changed, deprecated or removed from `prettyprinter`. How would a given release of `optparse-applicative` maintain a stable interface?

lehins:
@Bodigrim I think `ansi-wl-pprint` is no longer maintained. It is deprecated on hackage (see ekmett/ansi-wl-pprint#26) in favor of `prettyprinter`

I think it only makes sense for the community to start switching to the `prettyprinter` everywhere. Naturally it would be best to minimize  breakage, but I am sure it will not be possible to completely avoid it.

HuwCampbell:
Thanks for the PRs John. It's always to have people put time into optparse-applicative.

There are a couple of orthogonal points here, @Bodigrim's opinion is well warranted in my view. It's very important to me that optparse is API stable as a lot of libraries and programs depend on it both in the open source world and I imagine commercially as well (breaking peoples' builds for their choice of CLI library is a bit rich).

So the questions are:

- Is changing the underlying pretty printing library going to improve library sufficiently to warrant a reasonably large breaking change?
- Do we like some or all of these adjustments to the help text?

After the original issue mentioning the prettyprinter library, I did also make a branch to see what it would be like, but overall, my impression was that it's probably not worth the switch. ansi-wl-pprint is well maintained and stable, and while annotations are a nice feature, it's a bit of a can of worms as to what and how we might think to annotate, whether the annotations API should be user extensible and exposed, and so on.

It would probably make a coloured renderer a bit easier and more declaritive if that was a direction we decided to take, but in reality if we wanted coloured output we could do that already.


With regards to the actual rendering changes: I'm quite sure we should rethink how we do the hang when the program name and subcommand names are long as squashing the text into the side really does make it ugly. Starting them fresh on a new line with an indent really would be an improvement. I've been considering how it can be done without expanding the number of preference modifiers (again).

The pretty printing is a bit of a balancing act. I do find what you've got more readable above, but it also seems a bit verbose – one of your examples above requires 5 lines for only 3 options.

I'll run through the changes in detail when I get some more time.

newhoggy:
> What specifically do annotations buy us in case of optparse-applicative in production setting? What kind of richer console output are we looking for?

I was thinking perhaps HTML help or man page support would be nice.  It's not clear that annotations are strictly necessary to do this, but I'm thinking in these cases we could use annotations to delay some of the rendering to a later phase so we can re-use more code.

newhoggy:
> With regards to the actual rendering changes: I'm quite sure we should rethink how we do the hang when the program name and subcommand names are long as squashing the text into the side really does make it ugly. Starting them fresh on a new line with an indent really would be an improvement. I've been considering how it can be done without expanding the number of preference modifiers (again).
> 
> The pretty printing is a bit of a balancing act. I do find what you've got more readable above, but it also seems a bit verbose – one of your examples above requires 5 lines for only 3 options.

I'm wondering if a good compromise would be to have instead of this:

```
[ ( --certificate-redeemer-file FILE
  | --certificate-redeemer-value JSON VALUE
  )
  --certificate-execution-units (INT, INT)
]
```

something like this:

```
[ (--certificate-redeemer-file FILE | --certificate-redeemer-value JSON VALUE)
  --certificate-execution-units (INT, INT)
]
```

The rule would be if the list of options has no children, then it can collapse consecutively.  In the above `(--certificate-redeemer-file FILE | --certificate-redeemer-value JSON VALUE)` would wrap, so  it would actually be four lines.

newhoggy:
Another perhaps out-there idea would be to use rainbow parens 😁 .

Bodigrim:
@lehins thanks, I did not know that `ansi-wl-pprint` is deprecated, this is a change of perspective indeed. Nevertheless it still works fine at least with current GHC HEAD (future 9.4), so there is plenty of time to figure out migration plan.

newhoggy:
I'm cool for this to take a long as it needs, since we're currently on a fork.

My hope is for the switch to happen at an appropriate and opportune time and to either improve the layout in the library or otherwise export some way for users to have more control of the layout externally.

Bodigrim:
Out of curiosity I checked this branch against a standard `tasty` test suite. 

Before:
```
Usage: prop-compiled [-p|--pattern PATTERN] [-t|--timeout DURATION]
                     [-l|--list-tests] [-j|--num-threads NUMBER] [-q|--quiet]
                     [--hide-successes] [--color never|always|auto]
                     [--ansi-tricks ARG] [--quickcheck-tests NUMBER]
                     [--quickcheck-replay SEED] [--quickcheck-show-replay]
                     [--quickcheck-max-size NUMBER]
                     [--quickcheck-max-ratio NUMBER] [--quickcheck-verbose]
                     [--quickcheck-shrinks NUMBER]
```

After:
```
Usage: prop-compiled
                       [-p|--pattern PATTERN]
                       [-t|--timeout DURATION]
                       [-l|--list-tests]
                       [-j|--num-threads NUMBER]
                       [-q|--quiet]
                       [--hide-successes]
                       [--color never|always|auto]
                       [--ansi-tricks ARG]
                       [--quickcheck-tests NUMBER]
                       [--quickcheck-replay SEED]
                       [--quickcheck-show-replay]
                       [--quickcheck-max-size NUMBER]
                       [--quickcheck-max-ratio NUMBER]
                       [--quickcheck-verbose]
                       [--quickcheck-shrinks NUMBER]
```

It's probably a matter of habit, but to be honest I like previous pretty-printing better. That's because I usually have a half-screen tall console window, so vertical space is quite precious. Ideally I'd prefer something even more condensed than currently, e. g., 

```
Usage: prop-compiled 
    [-p|--pattern PATTERN] [-t|--timeout DURATION] [-l|--list-tests] 
    [-j|--num-threads NUMBER] [-q|--quiet] [--hide-successes] 
    [--color never|always|auto] [--ansi-tricks ARG] [--quickcheck-tests NUMBER] 
    [--quickcheck-replay SEED] [--quickcheck-show-replay] 
    [--quickcheck-max-size NUMBER] [--quickcheck-max-ratio NUMBER] 
    [--quickcheck-verbose] [--quickcheck-shrinks NUMBER]
```

Maybe a choice of layout could be configurable to a certain extent?..

newhoggy:
Thanks @Bodigrim for trying it out.

Some configurability would be nice.

There's also the possibility of choosing different layouts based on heuristics.  For example if the parser is highly nested, it starts to get confusing so a structured layout is preferable, whereas if the parser is mostly linear, then compact is usually better.

The came CLI tool may have the same multiple commands with different levels of complexity so a heuristic based layout selector would actually be quite nice.

lehins:
@Bodigrim I personally find this horrendous:

```
Usage: prop-compiled 
    [-p|--pattern PATTERN] [-t|--timeout DURATION] [-l|--list-tests] 
    [-j|--num-threads NUMBER] [-q|--quiet] [--hide-successes] 
    [--color never|always|auto] [--ansi-tricks ARG] [--quickcheck-tests NUMBER] 
    [--quickcheck-replay SEED] [--quickcheck-show-replay] 
    [--quickcheck-max-size NUMBER] [--quickcheck-max-ratio NUMBER] 
    [--quickcheck-verbose] [--quickcheck-shrinks NUMBER]
```

While this I personally find very eye appealing and much more readable as well:
```
Usage: prop-compiled
                       [-p|--pattern PATTERN]
                       [-t|--timeout DURATION]
                       [-l|--list-tests]
                       [-j|--num-threads NUMBER]
                       [-q|--quiet]
                       [--hide-successes]
                       [--color never|always|auto]
                       [--ansi-tricks ARG]
                       [--quickcheck-tests NUMBER]
                       [--quickcheck-replay SEED]
                       [--quickcheck-show-replay]
                       [--quickcheck-max-size NUMBER]
                       [--quickcheck-max-ratio NUMBER]
                       [--quickcheck-verbose]
                       [--quickcheck-shrinks NUMBER]
```

So this is quite a subjective matter and I do agree that an option to configure would be awesome! In fact, if that could be a user supplied option this would be even better ;) 

>  That's because I usually have a half-screen tall console window, so vertical space is quite precious"

I find it hard to believe that this is a widespread setup, because monitors are all widescreen nowadays  and splitting vertically is more efficient. That's why 80 char wrapping is still prevalent while coding

newhoggy:
I was considering, whether with `prettyprinter` annotations we could sprinkle some layout hint annotations in the document and postpone final layout decisions to the end where it is much easier to configure.

HuwCampbell:
Thanks all for the input.

There's no _best_ way to pretty print options and arguments in all situations for all programs, and I think this thread is a bit too opinionated. It's not worth the stress.

With regards to _this_ PR: it's too big for me to merge as is, it changes the underlying dependency tree as well as changing the rendering in ways which are at the very least controversial.

Here's my suggestion:
- We look at my [overflow](https://github.com/pcapriotti/optparse-applicative/tree/topic/overflow) branch to fix the immediate issue or the options being too squished up when there are long command names;
- We refocus this PR to just the pretty printer change and _not_ change the actual look; and
- We raise an _issue_ (not a PR) to discuss other changes to pretty printing.

That sound cool?

Huw

newhoggy:
Your suggestion is perfect.  I have been working on ANSI colour support, which will be necessary to do the switch because `optparse-applicative` exports such functions.  I can submit a PR to include both the library switch and the ANSI work.

newhoggy:
I have pushed an update to the PR as you've requested @HuwCampbell

andreasabel:
Since `ansi-wl-pprint` is deprecated in favor of `prettyprinter`, it makes sense to switch to the latter.  I recently researched a bit the alternatives for pretty printers, and came to the conclusion that `prettyprinter` would be my tool of choice.

optparse-applicative.cabal
```diff
@@ -87,19 +87,23 @@ library
...
                      , Options.Applicative.Help.Pretty
+                     , Options.Applicative.Help.Style
                      , Options.Applicative.Help.Types
                      , Options.Applicative.NonEmpty
                      , Options.Applicative.Types
                      , Options.Applicative.Internal
 
   build-depends:       base                            == 4.*
+                     , ansi-terminal                   >= 0.4.0
+                     , text                            >= 1.2
```
Bodigrim, line 103: Any chance to avoid `text` dependency? Otherwise tests/benchmarks inside `text` (and any packages it depends on) become impossible because of a circular dependency. Note that `prettyprinter` has a flag to depend on `base` only.

src/Options/Applicative/Help/Types.hs
```diff
@@ -42,6 +53,44 @@ helpText (ParserHelp e s h u d b g f) =
...
+
+renderCtxDecorated
+    :: Monoid ann
+    => ann
+    -> (Text -> B.Builder) -- ^ Render plain 'Text'
+    -> (ann -> ann -> B.Builder)  -- ^ How to render an annotation
+    -> (ann -> ann -> B.Builder)  -- ^ How to render the removed annotation
+    -> SimpleDocStream ann
+    -> B.Builder
+renderCtxDecorated topAnn toText push pop = go [topAnn]
```
Bodigrim, line 82: Is it possible to express `renderCtxDecorated` as `SimpleDocStream ann -> SimpleDocStream ()` pass, without touching `Data.Text.Builder` or anything else in `Data.Text` explicitly? I imagine you can replace `SAnn{Push,Pop}` with `SText` nodes, and then call `renderPretty` as before.

Otherwise (besides my primary interest to avoid `text` dependency) rendering `SimpleDocStream` on your own means a pretty fragile dependency on internals and semantics of `prettyprinter`, like the exhaustive list of constructors of `SimpleDocStream`, `panicUncaughtFail`, `panicInputNotFullyConsumed`, `panicUnpairedPop`, etc.

.hlint.template.yaml
```diff
@@ -0,0 +1,70 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
```
andreasabel, line 6: Why is this template file included in the PR?

out.html
```diff
@@ -0,0 +1 @@
+zsh: command not found: transaction
```
andreasabel, line 1: What is the role of this file?

src/Options/Applicative/Types.hs
```diff
@@ -109,25 +109,26 @@ data Backtracking
...
+                                                  -- subcommand fails (default: Backtrack)
+  , prefColumns :: Int                            -- ^ number of columns in the terminal, used to
+                                                  -- format the help page (default: 80)
+  , prefHelpLongEquals :: Bool                    -- ^ when displaying long names in usage and help,
+                                                  -- use an '=' sign for long names, rather than a
+                                                  -- single space (default: False)
+  , prefHelpShowGlobal :: Bool                    -- ^ when displaying subparsers' usage help,
+                                                  -- show parent options under a "global options"
+                                                  -- section (default: True)
+  , prefTabulateFill ::Int                        -- ^ Indentation width for tables
```
andreasabel, line 129: Space after `::` missing.

src/Options/Applicative/Types.hs
```diff
@@ -109,25 +109,26 @@ data Backtracking
...
-  , prefHelpLongEquals :: Bool    -- ^ when displaying long names in usage and help,
-                                  -- use an '=' sign for long names, rather than a
-                                  -- single space (default: False)
-  , prefHelpShowGlobal :: Bool    -- ^ when displaying subparsers' usage help,
-                                  -- show parent options under a "global options"
-                                  -- section (default: True)
-  , prefTabulateFill ::Int       -- ^ Indentation width for tables
-  } deriving (Eq, Show)
+  { prefMultiSuffix :: String                     -- ^ metavar suffix for multiple options
+  , prefDisambiguate :: Bool                      -- ^ automatically disambiguate abbreviations
```
andreasabel, line 113: How about, instead of increading the indentation, put the haddock on a new line, e.g.
```haskell
  ...
  , prefDisambiguate :: Bool
      -- ^ automatically disambiguate abbreviations (default: 'False')
  ...
  , foo :: Bar
      -- ^ haddocumentation for foo
  ... 
```

src/Options/Applicative/Help/Style.hs
```diff
@@ -0,0 +1,124 @@
...
+        [ if r then Just ANSI.Reset else Nothing
+        , fmap (\(intensity, c) -> ANSI.SetColor ANSI.Foreground intensity (convertColor c)) fg
+        , fmap (\(intensity, c) -> ANSI.SetColor ANSI.Background intensity (convertColor c)) bg
+        , fmap ANSI.SetConsoleIntensity b
+        , fmap (ANSI.SetItalicized . isItalicised) i
+        , fmap ANSI.SetUnderlining u
+        ]
+
+    convertColor :: Color -> ANSI.Color
+    convertColor = \c -> case c of
```
andreasabel, line 70: Use `\case`? (Since GHC 7.6)
```haskell
convertColor = \case
  Black -> ...
  ...
```

andreasabel (changes requested):
Just some comments on the form (not the content).

src/Options/Applicative/Help/Types.hs
```diff
@@ -42,6 +53,44 @@ helpText (ParserHelp e s h u d b g f) =
...
+    go (_:_:_)                SEmpty              = panicInputNotFullyConsumed
+    go (_:_)                  SEmpty              = mempty
+    go stack                  (SChar c rest)      = toText (T.singleton c) <> go stack rest
+    go stack                  (SText _l t rest)   = toText t <> go stack rest
+    go stack                  (SLine i rest)      = toText (T.singleton '\n') <> toText (textSpaces i) <> go stack rest
+    go stack@(ctxAnn:_)       (SAnnPush ann rest) = push ctxAnn ann <> go ((ctxAnn <> ann) : stack) rest
+    go (ann:stack@(ctxAnn:_)) (SAnnPop rest)      = pop ctxAnn ann <> go stack rest
+    go []                     SAnnPop{}           = panicUnpairedPop
+    go _                      (SAnnPush _ _)      = error "Unpaired push"
+    go _                      (SAnnPop _)         = error "Unpaired pop"
```
HuwCampbell, line 95: Sorry, I can't accept something like this function.

We really can't be doing full pattern matches on the internals of dependant libraries in something which sits close to the bottom of most applications dependency trees.

src/Options/Applicative/Help/Pretty.hs
```diff
@@ -39,10 +103,25 @@ ifNotAtRoot f doc =
...
+          PPI.Union x _       -> flatten x
+          PPI.Column f        -> PPI.Column (flatten . f)
+          PPI.WithPageWidth f -> PPI.WithPageWidth (flatten . f)
+          PPI.Nesting f       -> PPI.Nesting (flatten . f)
+          PPI.Annotated ann x -> PPI.Annotated ann (flatten x)
 
+          x@PPI.Fail   -> x
+          x@PPI.Empty  -> x
+          x@PPI.Char{} -> x
+          x@PPI.Text{} -> x
```
HuwCampbell, line 124: Why is this inlined?

src/Options/Applicative/Help/Core.hs
```diff
@@ -115,8 +121,8 @@ missingDesc = briefDesc' False
 -- | Generate a brief help text for a parser, allowing the specification
 --   of if optional arguments are show.
 briefDesc' :: Bool -> ParserPrefs -> Parser a -> Chunk Doc
-briefDesc' showOptional pprefs =
-  wrapOver NoDefault MaybeRequired
+briefDesc' showOptional pprefs = id
+    . wrapOver NoDefault MaybeRequired
```
HuwCampbell, line 125: why compose the identity function?

src/Options/Applicative/Help/Style.hs
```diff
@@ -0,0 +1,124 @@
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
```
HuwCampbell, line 2: Are these required?

src/Options/Applicative/Help/Style.hs
HuwCampbell: I would prefer to keep extensions low for this library.

.hlint.template.yaml
newhoggy: Removed.  Thanks!

out.html
newhoggy: Removed.  Thanks!

src/Options/Applicative/Types.hs
newhoggy: Fixed.  Thanks!

src/Options/Applicative/Types.hs
newhoggy: Done!

src/Options/Applicative/Help/Style.hs
newhoggy: Yeah, as much as I prefer using the extension I've avoided it to keep in line with existing code.

src/Options/Applicative/Help/Types.hs
newhoggy: I would like to suggest this isn't dependent on internals of libraries because the `prettyprinter` library has tutorials showing describing code very much like this which means to me this interface is public.

For example: https://github.com/quchen/prettyprinter/blob/9d7a40f2be5ce9a8e9c993b683e01d76eb06a079/prettyprinter/src/Prettyprinter/Render/Tutorials/StackMachineTutorial.hs#L99

newhoggy: Perhaps @quchen can comment?

src/Options/Applicative/Help/Core.hs
newhoggy: Stylistic thing to make all the lines consistent at the cost of using `id` which I expect to be optimised away by the component.  I've removed.

src/Options/Applicative/Help/Style.hs
newhoggy: Seems not.  Removed.  Thanks!

src/Options/Applicative/Help/Pretty.hs
newhoggy: What did you mean by inlined?

src/Options/Applicative/Help/Types.hs
newhoggy: This is a bit tricky.  I'm heading over to `prettyprinter` repo for advice.

src/Options/Applicative/Help/Types.hs
```diff
@@ -42,6 +48,43 @@ helpText (ParserHelp e s h u d b g f) =
...
+    go (_:_:_)                SEmpty              = SEmpty
+    go (_:_)                  SEmpty              = SEmpty
+    go stack                  (SChar c rest)      = SChar c (go stack rest)
+    go stack                  (SText l t rest)    = SText l t (go stack rest)
+    go stack                  (SLine i rest)      = SLine i (go stack rest)
+    go stack@(ctxAnn:_)       (SAnnPush ann rest) = push ctxAnn ann (go ((ctxAnn <> ann) : stack) rest)
+    go (ann:stack@(ctxAnn:_)) (SAnnPop rest)      = pop ctxAnn ann (go stack rest)
+    go _                      (SAnnPush _ _)      = error "An unpaired style initiator was encountered. This is a bug in the layout algorithm.  Please report this as a bug"
+    go _                      (SAnnPop _)         = error "An unpaired style terminator was encountered. This is a bug in the layout algorithm.  Please report this as a bug"
+{-# INLINE renderCtxDecorated #-}
```
newhoggy, line 90: This code has been updated with @Bodigrim 's suggestion to avoid using `text` and it looks a bit simpler.

src/Options/Applicative/Help/Types.hs
newhoggy: This code has been updated with @Bodigrim 's suggestion to avoid using `text` and it looks a bit simpler.

optparse-applicative.cabal
newhoggy: Done!

hasufell:
I tried this on ghcup.

Before:

```
Usage: ghcup [(-v|--verbose) | --no-verbose] [(-c|--cache) | --no-cache]
             [--verify | (-n|--no-verify)] [(-o|--offline) | --no-offline]
             [--gpg <strict|lax|none>] COMMAND
```

After:

```
Usage: ghcup [(-v|--verbose) | --no-verbose]
               [(-c|--cache) | --no-cache]
               [--verify | (-n|--no-verify)]
               [(-o|--offline) | --no-offline]
               [--gpg <strict|lax|none>]
               COMMAND
```

It looks a little bit off-balance.

newhoggy:
Would it help if I break up this PR to be only the switch to the `prettyprinter` library and have any stylistic changes in a separate PR?

src/Options/Applicative/Help/Pretty.hs
newhoggy: `flatten` came from the deprecated `ansi-wl-pprint` library's internal module.  `prettyprinter` doesn't export such a function.

The function is defined in the `where` clause because this is the only function that needs to use it.

newhoggy:
I've removed all the code that is unrelated to the switch to `prettyprinter`.

andreasabel (approved):
I second the switch to `prettyprinter`.  
I think the breakage caused by this is mitigated by the fact that `optparse-applicative` exports `Doc` and its combinators.

Bodigrim (approved):
Looks good from my perspective, I can still run `bytestring` and `text` test suite without cyclical dependency.

Jimbo4350:
Is this PR good to be merged? Is there anything else that needs to be done? @HuwCampbell @andreasabel @Bodigrim

andreasabel:
CI hasn't run yet.  @HuwCampbell , could you please hit the button?

HuwCampbell:
I've set this to run in CI.

So my only real issue is that this doesn't really _do_ that much. The focus here was to be as backwards compatible as possible which is good, but it's inlining a few functions which aren't exported, and really only sticking to what we can do at the moment.

I have a little branch I'm working on as well (indeed I've had a look a few times), but I think if we want to make a breaking change it should offer more real advantages.

My current thought is that we should probably use the annotations for a bit more than what is currently available, for example, optionally making the `|` in alternatives duller when desired.

I know it's not super satisfying, but I don't want to push two big breaking changes to the pretty-printer.

I'll set some time in my calendar for September to finish it off and figure out a nice API for user preference.

johnk-da:
I pushed a fix for older compilers:

* Removed use of `TypeApplications`
* Import `Data.Semigroup`

newhoggy:
Any progress on moving to `prettyprinter`?

CarlosLopezDeLara:
Any updates on this one? Are there still any concerns? Thanks

HuwCampbell:
Thank you for putting in this work, I used a bunch of the ideas in a recent release which achieves these goals.

newhoggy:
Thanks @HuwCampbell for putting together the changes for the new release.

PR #436, opened 2021-08-16
Haddock for fullDesc: point out that this is the default

andreasabel:
Haddockumentation for `fullDesc`: point out that this alternative is the default one.


HuwCampbell:
Ta

PR #438, opened 2021-09-14
Added support for simple arguments through Generics

chisui:
This adds `simpleOptions` and `simpleCommands` which create a `Parser` for a type through `GHC.Generics`.

* `simpleOptions` expects the data type to have a single constructor. Each record is interpreted as an option. A `Bool` record results in a `switch`, a `String` in a `strOption` and any other type in a `option auto`. Constructors without records result in series of arguments.
* `simpleCommands` takes a data type with multiple constructors and creates a command for each of them. The constructors themselves are handled the same as for `simpleOptions`

Exampltes can be seen [here](https://github.com/chisui/optparse-applicative/blob/master/tests/Examples/Generic/Simple.hs) and [here](https://github.com/chisui/optparse-applicative/blob/master/tests/Examples/Generic/Commands.hs).

---

This is a proof of concept, so I haven't added haddock or documentation yet. The code also doesn't match the coding style of the rest of the code. If this is something that could be added to aptparse-applicative I will polish the PR.

Theoretically this could be handled by a separate library, but I would rather have this as part of the main lib, since it's just one file. Although I have only tested this with `ghc 8.10.7` and don't know about the compatibility of the type level language extensions used in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ cabal.sandbox.config
 
 .stack-work/
 stack.yaml.lock
+
+.envrc
+.history
+
--- a/nix/build.nix
+++ b/nix/build.nix
@@ -1,0 +1,5 @@
+{ pkgs ? import ./nixpkgs.pinned.nix 
+}: let
+  hPkgs = pkgs.haskell.packages.ghc8107;
+in { inherit hPkgs; } // hPkgs.callCabal2nix "optparse-applicative" ../. {}
+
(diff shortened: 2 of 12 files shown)


HuwCampbell:
Interesting. I haven't really considered doing this here because the [optparse-generic](https://hackage.haskell.org/package/optparse-generic) library seems to handle it quite well.

I'll take a look at your implementation in more detail after work.

chisui:
I didn't know about optparse-generic. It does what I tried to do in a more elegant way.

PR #441, opened 2021-11-02
runParser: Throws a ParseError that contains the parser

teto:
when running live completion (e.g., with haskeline), it allows to pinpoint which parser failed autocompletion. As a follow up, one can retreive completions for that specific parser and display those to the user


HuwCampbell:
This breaks error messages when using the parser (not doing autocompletions). From the test suite:

`commands_header -zello`

```
Invalid option `-zello'

Did you mean this?
    hello

Usage: commands_header COMMAND
```
/=
```
Missing: COMMAND

Usage: commands_header COMMAND
```



The parser is actually still embedded in the

PR #447, opened 2022-02-04
runParser: Throws a ParseError that contains the parser

teto:
the current completers are ok for shell completion, but when dealing with live haskeline completion, I need to get the possible candidates when the command fails. It's not possible with parseError so I replaced it with exitP that returns a parser in the error so in my code I can list the possible completions:
https://github.com/teto/mptcpanalyzer/blob/7bc09ea02926a9e99249c3833e0ae251980babc9/mptcpanalyzer/src/MptcpAnalyzer/Utils/Completion.hs#L111

Alternatively, more constructors of ParseError should contain a SomeParser to help the completer list the available completions in case of an error. Let me know which you favor.

### DIFF
--- a/src/Options/Applicative/Common.hs
+++ b/src/Options/Applicative/Common.hs
@@ -207,7 +207,7 @@ runParser policy isCmdStart p args = case args of
   (arg : argt) -> do
     (mp', args') <- do_step arg argt
     case mp' of
-      Nothing -> hoistMaybe result <|> parseError arg p
+      Nothing -> hoistMaybe result <|> exitP isCmdStart policy p Nothing
       Just p' -> runParser (newPolicy arg) CmdCont p' args'
   where
     result =
--- a/tests/commands_header.err.txt
+++ b/tests/commands_header.err.txt
@@ -1,6 +1,3 @@
-Invalid option `-zello'
-
-Did you mean this?
-    hello
+Missing: COMMAND
 
 Usage: commands_header COMMAND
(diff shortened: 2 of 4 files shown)


HuwCampbell:
I'm not sure what you're suggesting I do here.

I think you changed the call from parseError to `exitP` so a `SomeParser` could be invoked, but the value given from parseError already contains a `SomeParser`.

Indeed, that's how we make the suggestions 

> Invalid option `-zello'
> 
> Did you mean this?
>     hello

The "hello" is derived from the parser at the point of the error.

```haskell
parseError :: MonadP m => String -> Parser x -> m a
parseError arg p = errorP $ UnexpectedError arg (SomeParser p)
```

Which constructor are you to `ParseError` are you referring to?

I'm interested in your use case but I can't merge this PR as it makes the usage text obviously worse for most users.

teto:
Right, sry I wrote the code some time ago and got confused as I am looking at it again.
I think that's the part that didn't return a parser https://github.com/teto/optparse-applicative/blob/ad5d43ad8f4cbca13ca4d0658926e81a2d1e7143/src/Options/Applicative/Common.hs#L190

I think I was thinking about this part actually, I will update the PR.
```
-      ArgReader rdr ->
-        fmap pure . lift . lift $ runReadM (crReader rdr) arg
-      _ -> mzero
+      ArgReader rdr -> let
+          tmp = runExcept $ runReaderT (unReadM $ crReader rdr) arg 
+        in
+        -- rewrite the error so that completion can access the parser
+        fmap pure . lift . lift $ either (const $ exitP CmdStart AllPositionals parser Nothing) return tmp
+      _ -> mzero) parser  
 ```

Issue #193, opened 2016-03-17
Commands are not disambiguated


deepfire:
Commands are not disambiguated (while options are):

```
params_parser ∷ Parser Params
params_parser =
    Params
    <$> strOption (long "server"  <> metavar "HOSTNAME" <> help "Server")
    <*> strOption (long "login"   <> metavar "LOGIN" <> help "Username")
    <*> hsubparser
            (command "branches"
             (info
              (Branches
               <$> subparser ((command "git"    (info (pure Git)    $ progDesc "Git branches")) <>
                              (command "jira"     (info (pure Jira)     $ progDesc "Jira branches"))))
              (fullDesc <> progDesc "Print and intersect git/jira branches")))
```

HuwCampbell:
Indeed. I don't think this is a very high priority, but I can see some might find it useful.

HuwCampbell:
This was fixed a while back.

PR #454, opened 2022-07-06
Allow commands to be disambiguated

HuwCampbell:
Fix for #193 .

### DIFF
--- a/src/Options/Applicative/Types.hs
+++ b/src/Options/Applicative/Types.hs
@@ -242,14 +242,14 @@ data OptReader a
   -- ^ flag reader
   | ArgReader (CReader a)
   -- ^ argument reader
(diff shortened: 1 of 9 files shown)


src/Options/Applicative/Types.hs
```diff
@@ -242,14 +242,14 @@ data OptReader a
   -- ^ flag reader
   | ArgReader (CReader a)
   -- ^ argument reader
-  | CmdReader (Maybe String) [String] (String -> Maybe (ParserInfo a))
+  | CmdReader (Maybe String) [(String, ParserInfo a)]
```
HuwCampbell, line 245: This is the main change.

The `lookup` function is now not embedded creating this type, but selected in the searchParser.

PR #455, opened 2022-08-05
Add --version parser

m1-s:
This MR contains a new `InfoMod` that allows to specify a version which can be requested with `-v` or `--version`.

Open questions:

- Is `version` a too generic name? Maybe `appVersion` is better?
- Does this need more documentation somewhere?
- Maybe not occupy `-v` as it may be used differently by some apps?

### DIFF
--- a/src/Options/Applicative.hs
+++ b/src/Options/Applicative.hs
@@ -74,6 +74,7 @@ module Options.Applicative (
   abortOption,
   infoOption,
   helper,
+  simpleVersioner,
 
   -- ** Modifiers
   --
--- a/src/Options/Applicative/Extra.hs
+++ b/src/Options/Applicative/Extra.hs
@@ -6,6 +6,7 @@ module Options.Applicative.Extra (
   helper,
   helperWith,
   hsubparser,
+  simpleVersioner,
   execParser,
   customExecParser,
   execParserPure,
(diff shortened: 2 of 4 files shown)


tfc:
@HuwCampbell i am also very interested in this feature as this is something that is duplicated practically across *all* apps ever.

HuwCampbell:
Thanks for opening this conversation.

So, obviously a lot of people add these, and really it's not that hard, but there are a few subtleties which I discuss in the [wiki](https://github.com/pcapriotti/optparse-applicative/wiki/Version-Options).

The way you've used an `InfoMod` is kind of nice, but out of step with `helper`. Maybe we could migrate to the helper function being just included with a modifier to not? Until then it's a bit janky to do things both ways.

Here's what I suggest in the wiki:
```
versioner :: Parser (a -> a)
versioner = infoOption (showVersion version) (long "version" <> help "Show version" <> hidden)

opts :: Parser Command
opts = info (base <**> helper <**> versioner) idm
```
 
As to your specific questions:

a) Yes, clearly, as naming it version even broke the test suite, and hundreds of people having to now do qualified imports and be annoyed at me would not be fun.
b) The wiki has some. Otherwise great haddocks and examples in the readme go a long way.
c) Tough call, and one of the reasons I haven't just added this already. `-v` is either "version", or "verbose" in standard conventionl while `-h` is almost universally "help". So one might want to make this an argument to the function.

The silly thing though is that as soon as we start adding customisations to versioner, we're almost writing the whole thing anyway!

I'll think about this more.

m1-s:
@HuwCampbell Have you had time to make your mind up yet? Is there anything I can do to move this forward?

m1-s:
@HuwCampbell We are really waiting for a solution here. Any news?

HuwCampbell:
I think for now:

call it `simpleVersioner` and implement in a similar way to helper.

```haskell
simpleVersioner :: String -> Parser (a -> a)
```

m1-s:
> I think for now:
> 
> call it `simpleVersioner` and implement in a similar way to helper.
> 
> ```haskell
> simpleVersioner :: String -> Parser (a -> a)
> ```
> 
> and skip the short option `-v` so we don't collide.

Done. Let me know if anything is missing.

m1-s:
Fixed build pipe line. Please run again

m1-s:
@HuwCampbell

HuwCampbell:
Thanks! Merged a while back.

PR #456, opened 2022-08-16
Update code snippet in parserFailure haddocks

jaspervdj:
ShowHelpText takes an extra argument now so the snippet was failing compilation.


HuwCampbell:
Thanks

PR #469, opened 2023-04-08
Put back example for `optional` in README.

trespaul:
Give example of usage with `optional`, as introduced by e7cc1568c5a112904d06837a086201de0431f6c8 but then removed by 8de333d2aeb6cf32261ff261db48d613b01c33c2. Its removal has caused a slight inconvenience I think for those not as well-versed in `Control.Applicative` combinators, e.g. [this SO question](https://stackoverflow.com/questions/32422339/how-to-parse-an-optional-flag-as-a-maybe-value).


HuwCampbell:
Hi, thanks, I'm on holidays for a bit so I'll look at this when I'm back.

HuwCampbell:
Ta

PR #471, opened 2023-05-05
feat: add optional argument builder

drlkf:
This is just a small helper that I've looked for in the docs and I haven't found. It seems like an elementary feature so there might be a reason why it wasn't included before, but I'm submitting just in case.


pcapriotti:
You can use `optional` from `Control.Applicative` to get the same effect. There is an example [here](https://github.com/pcapriotti/optparse-applicative#alternative) in the docs.

drlkf:
Right, aren't I silly. Thanks !

PR #473, opened 2023-05-19
Bump pretty printer and stop relying on its internal modules.

HuwCampbell:
A recent release of ansi-wl-pprint just changed it to reexport prettyprinter's compat module. By supporting this bump, it unblocks migration of downstream packages.

This means that the prettyprinter library can now be used instead of the ansi-wl-pprint library depending on the version bounds. This was a bit tricky to get right, as there were a few tricks we pulled using the internal modules. Figured it out though.

I've turned off deprection warnings on the module, which is a bit savage, as otherwise it emits many, many, warnings.

This is a half way step, but an acceptable one in my opinion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+- Widen bounds for `ansi-wl-pprint`. This supports the use of `prettyprinter`
+  in a non-breaking way, as the `ansi-wl-pprint > 1.0` support the newer
+  library.
+
 - Export `helpIndent` from `Options.Applicative`.
 
 - Export completion script generators from `Options.Applicative.BashCompletion`.
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -103,7 +103,7 @@ library
   build-depends:       base                            == 4.*
                      , transformers                    >= 0.2 && < 0.7
                      , transformers-compat             >= 0.3 && < 0.8
-                     , ansi-wl-pprint                  >= 0.6.8 && < 0.7
+                     , ansi-wl-pprint                  >= 0.6.8 && < 1.1
 
   if flag(process)
     build-depends:     process                         >= 1.0 && < 1.7
--- a/src/Options/Applicative/BashCompletion.hs
+++ b/src/Options/Applicative/BashCompletion.hs
@@ -1,10 +1,12 @@
+{-# OPTIONS_GHC -fno-warn-warnings-deprecations #-}
 -- | You don't need to import this module to enable bash completion.
 --
 -- See
 -- <http://github.com/pcapriotti/optparse-applicative/wiki/Bash-Completion the wiki>
 -- for more information on bash completion.
 module Options.Applicative.BashCompletion
-  ( bashCompletionParser, 
+  ( bashCompletionParser,
+
     bashCompletionScript,
     fishCompletionScript,
     zshCompletionScript,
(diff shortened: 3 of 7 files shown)


TravisCardwell:
I experimented with building one of my projects using this branch, just to get an idea of what will be involved to upgrade.  The only issue that came up is that the [`CmdReader` constructor][] of [`OptReader`][] has been changed.  Investigating, it was changed as part of the [Allow disambiguation to work with commands as well as options][] commit.

[`CmdReader` constructor]: <https://github.com/pcapriotti/optparse-applicative/blob/50e0f8f53da302a28c6b9cabec77887e644b7828/src/Options/Applicative/Types.hs#L245-L246>
[`OptReader`]: <https://hackage.haskell.org/package/optparse-applicative-0.17.0.0/docs/Options-Applicative-Types.html#t:OptReader>
[Allow disambiguation to work with commands as well as options]: <https://github.com/pcapriotti/optparse-applicative/commit/6f74f3be4e4a9571feb8aedeb5efed87f3352aaa>

```
$ git diff d307e46 50e0f8f src/Options/Applicative/Types.hs
```

```diff
diff --git a/src/Options/Applicative/Types.hs b/src/Options/Applicative/Types.hs
index 0644676..a556f2a 100644
--- a/src/Options/Applicative/Types.hs
+++ b/src/Options/Applicative/Types.hs
@@ -242,14 +242,14 @@ data OptReader a
   -- ^ flag reader
   | ArgReader (CReader a)
   -- ^ argument reader
-  | CmdReader (Maybe String) [String] (String -> Maybe (ParserInfo a))
+  | CmdReader (Maybe String) [(String, ParserInfo a)]
   -- ^ command reader

 instance Functor OptReader where
   fmap f (OptReader ns cr e) = OptReader ns (fmap f cr) e
   fmap f (FlagReader ns x) = FlagReader ns (f x)
   fmap f (ArgReader cr) = ArgReader (fmap f cr)
-  fmap f (CmdReader n cs g) = CmdReader n cs ((fmap . fmap) f . g)
+  fmap f (CmdReader n cs) = CmdReader n ((fmap . fmap . fmap) f cs)

 -- | A @Parser a@ is an option parser returning a value of type 'a'.
 data Parser a
```

At least with my usage, it is easy to change the code to match the new API.  CPP is required, however, in order to be compatible with older and well as newer versions of the library.

I am mentioning this because the constructor change is not yet in the changelog.

HuwCampbell:
Yes. Thanks.

I've got a separate tag for https://github.com/pcapriotti/optparse-applicative/tree/patch/0.17.1 which will be the patch branch.

HuwCampbell:
Out of curiosity, what were you using the constructors for?

I was pretty careful to not change the API in the main module to support this functionality, so I'm a bit surprised it affected you.

I'm aware it's a PVP break though, so will only go into 0.18.

TravisCardwell:
When working on executables that use many hierarchical commands, the top-level `--help` displays the commands available, and I set the `progDesc` of each command to display the sub-commands.  This gives users a broader overview, helping them avoid having to perform `--help` on each command when searching or exploring.  To implement this, I have used the following function to get the names of commands within a command parser.  I only need to update this function for compatibility with the new constructor, and it is an easy change.

```haskell
commands :: OA.Parser a -> [String]
commands =
    let go _ opt = case OAT.optMain opt of
           OAT.CmdReader _ cmds _ -> reverse cmds
           _otherReader           -> []
    in  concat . OAC.mapParser go
```

PR #474, opened 2023-05-20
Move to Pretty Printer Proper

HuwCampbell:
This is a proposed 0.18 release with the Prettyprinter support proper.

We will probably need to drop some older GHC versions for this, but
they're _really_ old now.

### DIFF
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -100,10 +100,11 @@ library
                      , Options.Applicative.Types
                      , Options.Applicative.Internal
 
-  build-depends:       base                            == 4.*
+  build-depends:       base                            >= 4.5 && < 5
                      , transformers                    >= 0.2 && < 0.7
                      , transformers-compat             >= 0.3 && < 0.8
-                     , ansi-wl-pprint                  >= 0.6.8 && < 1.1
+                     , prettyprinter                   >= 1.7 && < 1.8
+                     , prettyprinter-ansi-terminal     >= 1.1 && < 1.2
 
   if flag(process)
     build-depends:     process                         >= 1.0 && < 1.7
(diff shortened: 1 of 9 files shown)


HuwCampbell:
@Bodigrim 

As you have some specific constraints for this library regarding the `text` package and this one, would you like to test whether this still suits your needs?

optparse-applicative.cabal
```diff
@@ -100,10 +100,11 @@ library
                      , Options.Applicative.Types
                      , Options.Applicative.Internal
 
-  build-depends:       base                            == 4.*
+  build-depends:       base                            >= 4.5 && < 5
                      , transformers                    >= 0.2 && < 0.7
                      , transformers-compat             >= 0.3 && < 0.8
-                     , ansi-wl-pprint                  >= 0.6.8 && < 0.7
+                     , prettyprinter                   >= 1.7 && < 1.8
+                     , prettyprinter-ansi-terminal     >= 1.1 && < 1.2
```
Bodigrim, line 107: `prettyprinter-ansi-terminal` is problematic for my use cases, because it depends on `text` unconditionally. 

Needless to say that my use cases are quite extreme and exotic. You are doing an excellent job in this PR.

Sorry, I don't remember, does `optparse-applicative` use any ANSI capabilities like colors or bold/italic?

HuwCampbell: Not directly. I didn't want to add an annotation type parameter to `Parser`, so went with one which is capable enough (and which would still support what downstream users might have in their doc strings).

So for your use-case we could make a later release where this import is behind a flag.

Bodigrim: This suggests that ideally `ParserInfo a` should be additionally parameterized by `ann` from `Doc ann`, but it's probably too much of a breakage for a stable package. 

Sorry, I don't have much time to delve into it right now, so I suggest you go ahead as is, and I'll try to figure out workarounds for my use case later. Thanks for remembering about my concerns, it's a pleasure to deal with you.

Issue #478, opened 2023-05-26
How do I color a command name?


modotte:
Hello! I'm having trouble finding a solution to this. Is something like this currently possible?

Currently, what I tried to so far -
```haskell
myCommands :: Parser CustomCommand
myCommands =
  O.subparser $
    O.command
      "myCommand"
      ( O.info
          O.empty
          (O.progDesc "description here")
      )
      <> O.style OH.green
```

but I got the "metavar" part of it colored green instead, not `myCommand` which is what I'm looking for:
![image](https://github.com/pcapriotti/optparse-applicative/assets/85026096/b6ccb6d3-2031-4336-8fbb-74b45b27443a)

HuwCampbell:
Sorry this isn't possible right now.

Martinsos:
Ah that is too bad!

Before optparse-applicative, our help looked like this:
![image](https://github.com/pcapriotti/optparse-applicative/assets/1536647/b9713aec-2a31-4709-b314-91bcc669a382)

but now, with optparse, it looks like this:

![image](https://github.com/pcapriotti/optparse-applicative/assets/1536647/e1899d8a-cd18-4332-9dd3-6f5a2f1aec41)

It is not terrible, but we had titles for the groups of commands bold, and commands where yellowish, and we would like to keep those styles.

Are there any considerations to add this in the future?
If we wanted to make a PR for it ourselves, could you recommend a place where to start, the direction?

HuwCampbell:
Yeah.

So one of the things I'm considering is adding semantic annotations to the pretty printer document which are then compiled to ANSI codes.

Thanks for sharing.

HuwCampbell:
Just a hint, the extra (Command | Command) can be fixed with hidden on one of them.

Martinsos:
> Yeah.
> 
> So one of the things I'm considering is adding semantic annotations to the pretty printer document which are then compiled to ANSI codes.
> 
> Thanks for sharing.

That sounds interesting.

I was hoping there would be some easy way to provide our own logic for "styling" the help output. For example, I could provide a function that takes subgroup name as an input and returns it stylized (well, surrounded with ANSI codes in my case but could be anything, emojis or something) and then that is used in its place. There would also be a function for command names.

HuwCampbell:
The way forward is to use a SemanticDoc type internally, which annotates the Doc tree with what each printed item is (like, a command, or optional flag).

The user facing API should stay the same.

[...]

PR #482, opened 2023-06-10
Allow customization of help info via introducing semantic Doc

Martinsos:
Should fix https://github.com/pcapriotti/optparse-applicative/issues/478 .

@HuwCampbell this is still more of a playground then a solution, and it doesn't even compile completely, but I think it is starting to take some initial shape, so I thought I would rather share while still in this early stage, so you can course-correct me if I am going in a very wrong direction.

I mostly asked questions in the code, in comments, about things that are not clear to me.

The approach is that I introduced a new type of annotation, `HelpAnn`, which still allows for AnsiStyle annotations (to support users defining it via modifiers, and also the custom docs they can already provide for some parts of docs like header, footer and similar), but also allows for `HelpType` annotations, which we use internally to annotate parts.

The idea is then to later, probably at SimpleDocStream stage, reannotate `HelpAnn` annotations into `AnsiStyle` annotations and continue from there same as before.

I haven't yet started at all working on modifiers that would allow users to specify custom style. For now, I just focused on replacing one kind of annotation with another kind of annotation. I mostly got stuck on figuring out where to exactly annotate what with our new semantic annotation, because I am having hard time figuring out which `Chunk Doc` is presenting what from the code.

Once we have internal semantic annotation working, and a default way defined of converting those annotations into ansi annotations (probably we will just do nothing for most of them), we can start looking into allowing users to define their own function that transforms `HelpAnn` into `AnsiStyle` as one way of customization, and style modifier as another way of customization.

Interested to hear what you think, does this direction make sense? Thanks!


HuwCampbell:
Hi.

Yes that's pretty much what I was suggesting. Please don't reformat everything in a PR though. It's too hard to tell what the point is.

Martinsos:
> Hi.
> 
> Yes that's pretty much what I was suggesting. Please don't reformat everything in a PR though. It's too hard to tell what the point is.

Ah sorry I didn't realize I was doing reformatting! That was ormolu doing it on its own. I will have to turn it off and then re-do the changes.
In the meantime, do you have any pointers? You said direction seems to be ok, that is great -> what about good places to annotate stuff? I annotated on about two places, but I am not sure about some other places where I should do it. It would help if you could comment on the TODOs that I left in the code.

src/Options/Applicative/Help/Core.hs
```diff
@@ -1,119 +1,133 @@
...
       case optMain opt of
         CmdReader gn cmds ->
           (,) gn $
-            tabulate (prefTabulateFill pprefs)
-              [ (pretty nm, align (extractChunk (infoProgDesc cmd)))
-              | (nm, cmd) <- reverse cmds
+            tabulate
+              (prefTabulateFill pprefs)
+              [ ( annotateHelp CmdName $ pretty cmdName,
+                  align (annotateHelp Description $ ansiDocToHelpDoc $ extractChunk (infoProgDesc cmdInfo))
```
Martinsos, line 113: I did some annotating here.

src/Options/Applicative/Help/Core.hs
```diff
@@ -204,65 +219,67 @@ optionsDesc global pprefs = tabulate (prefTabulateFill pprefs) . catMaybes . map
...
+      with_title (fromMaybe "Available commands:" groupName) $
+        vcatChunks (snd <$> cmdGroup)
+    cmdGroupHelp _ = mempty
 
-    with_title :: String -> Chunk Doc -> Chunk Doc
-    with_title title = fmap (pretty title .$.)
+    cmdGroups = groupBy ((==) `on` fst) $ cmdDesc pprefs p
 
+    with_title :: String -> Chunk HelpDoc -> Chunk HelpDoc
+    with_title title = fmap (annotateHelp Title . (pretty title .$.))
```
Martinsos, line 272: I did a bit of annotating here.

src/Options/Applicative/Help/Core.hs
```diff
@@ -37,27 +37,37 @@ import Options.Applicative.Common
...
+optDesc :: ParserPrefs -> OptDescStyle -> ArgumentReachability -> Option a -> (Chunk HelpDoc, Parenthetic)
 optDesc pprefs style _reachability opt =
   let names =
         sort . optionNames . optMain $ opt
       meta =
-        stringChunk $ optMetaVar opt
+        annotateHelp Metavar <$> stringChunk (optMetaVar opt)
       descs =
-        map (pretty . showOption) names
+        map (annotateHelp OptionName . pretty . showOption) names
```
Martinsos, line 70: I did some annotating here.

src/Options/Applicative/Help/Core.hs
```diff
@@ -88,32 +98,38 @@ optDesc pprefs style _reachability opt =
...
       case optMain opt of
         CmdReader gn cmds ->
           (,) gn $
             tabulate (prefTabulateFill pprefs)
-              [ (pretty nm, align (extractChunk (infoProgDesc cmd)))
-              | (nm, cmd) <- reverse cmds
+              [ (annotateHelp CmdName $ pretty cmdName,
+                 align (annotateHelp Description $ ansiDocToHelpDoc $ extractChunk (infoProgDesc cmdInfo))
+                )
+              | (cmdName, cmdInfo) <- reverse cmds
```
Martinsos, line 117: Some more annotating here.

src/Options/Applicative/Types.hs
```diff
@@ -90,9 +91,9 @@ data ParserInfo a = ParserInfo
   { infoParser :: Parser a    -- ^ the option parser for the program
   , infoFullDesc :: Bool      -- ^ whether the help text should contain full
                               -- documentation
-  , infoProgDesc :: Chunk Doc -- ^ brief parser description
-  , infoHeader :: Chunk Doc   -- ^ header of the full parser description
-  , infoFooter :: Chunk Doc   -- ^ footer of the full parser description
+  , infoProgDesc :: Chunk AnsiDoc -- ^ brief parser description
+  , infoHeader :: Chunk AnsiDoc   -- ^ header of the full parser description
+  , infoFooter :: Chunk AnsiDoc   -- ^ footer of the full parser description
```
Martinsos, line 96: This is still AnsiDoc, so that existing interface or these definitions doesn't change and so that users can continue using ansi styles while defining these.

src/Options/Applicative/Types.hs
```diff
@@ -150,11 +151,11 @@ data OptVisibility
 -- | Specification for an individual parser option.
 data OptProperties = OptProperties
   { propVisibility :: OptVisibility       -- ^ whether this flag is shown in the brief description
-  , propHelp :: Chunk Doc                 -- ^ help text for this option
+  , propHelp :: Chunk HelpDoc                 -- ^ help text for this option
   , propMetaVar :: String                 -- ^ metavariable for this option
   , propShowDefault :: Maybe String       -- ^ what to show in the help text as the default
   , propShowGlobal :: Bool                -- ^ whether the option is presented in global options text
-  , propDescMod :: Maybe ( Doc -> Doc )   -- ^ a function to run over the brief description
+  , propDescMod :: Maybe ( HelpDoc -> HelpDoc )   -- ^ a function to run over the brief description
```
Martinsos, line 158: I am not quite sure what these are, therefore also not quite sure if these should be HelpDoc or AnsiDoc.

Martinsos: I am not quite sure if these should operate on HelpDoc or AnsiDoc?

Martinsos:
@HuwCampbell I reverted the formatting changes, so now it should be ready for a closer look!

src/Options/Applicative/Help/Chunk.hs
```diff
@@ -22,6 +22,14 @@ import Prelude
...
 import Options.Applicative.Help.Pretty
 
+-- TODO: What is the point of Chunk in this codebase? Is it basically used to give Doc additional powers -> Monoid?
+--   I wonder if it wasn't better if it was named DocPlus or Doc' or smth like that, and also a newtype, and then
+--   used like that around: `newtype DocPlus a = DocPlus (Chunk (Doc a))`, or maybe we can kick `Chunk` out completely?
+--   Feels to me like "Chunk" abstraction is a bit confusing: how does it relate to Doc? Is it
+--   really a "doc chunk"? But isn't Doc already a "chunk of doc"?.
+
+-- TODO: We have two types of functions here: general (Chunk a) operations, and Chunk (Doc a) operations. We should probably split those into separate modules.
+
```
HuwCampbell, line 32: It's so we can get a different monoid instance and not add spaces or breaks when things are hidden.

You'll find if its removed that internal and hidden options will affect the way the parser is rendered.

HuwCampbell:
I spent 10 minutes on the train and got it to this:
image

HuwCampbell:
All up your approach it good. Do you mind if I push some commits to your branch?

Martinsos:
That's looking great! @HuwCampbell sure pls go for it -> I am currently slammed with other work, but might be able to dedicate some more time in a couple of weeks. But if you can make progress / get it done in the meantime, none happier than me!

Btw this is what we currently have in our CLI, and what we would like to also achieve with optparse applicative:

image

So I am really hoping to achieve level of customizability that can bring us close to that.

HuwCampbell:
Will do. We're still a way to go though.

The issue isn't actually providing colours per se, that's relatively easy. I also want to respect if it's a TTY; environment variables like `NOCOLOR`; and a `--no-color` flag – while also giving a good programming API.

I general, if it's not a TTY, one shouldn't print colours; same for the other two ways of suppressing them. This gets a bit tricky for us because at the moment we don't have an environment parser to lean on internally.

Martinsos:
> Will do. We're still a way to go though.
> 
> The issue isn't actually providing colours per se, that's relatively easy. I also want to respect if it's a TTY; environment variables like `NOCOLOR`; and a `--no-color` flag – while also giving a good programming API.
> 
> I general, if it's not a TTY, one shouldn't print colours; same for the other two ways of suppressing them. This gets a bit tricky for us because at the moment we don't have an environment parser to lean on internally.

Ah interesting, I completely forgot about that. So we are really talking about the support for the ansi escape codes? And you are saying we don't have any logic for checking that at the moment + for propagating it to the right places?

PR #487, opened 2024-05-21
Fix broken Haddock links

pgujjula:
Fix all broken Haddock links. Now Haddock doesn't emit any warnings about out-of-scope identifiers.


HuwCampbell:
Ta

PR #488, opened 2024-05-24
[feat] allow passing list literals to mod builders

MangoIV:
I have found myself writing things like `strOption $ mconcat [... ` quite often and upon discussion, @Qqwy suggested just using list literals, this PR implements this idea, it works like a charm. the `IsList` instance even respects its laws!

### DIFF
--- a/src/Options/Applicative/Builder.hs
+++ b/src/Options/Applicative/Builder.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE TypeFamilies #-}
 module Options.Applicative.Builder (
   -- * Parser builders
   --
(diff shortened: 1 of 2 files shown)


HuwCampbell:
That's an interesting approach.

I try to keep optparse-applicative very conservative with regards to its extensions (we're just Haskell 98 + Existential Types at the moment), so I don't think it's very likely this will be merged as is.

But I'll keep this open now so it can be discussed further.

MangoIV:
well the extension is required to write the instance, it’s not that I’m making excessive use of TypeFamilies, unfortunately there’s no other good place for this to go, orphan instances are sad :( 

Alright, let’s see :)

MangoIV:
@HuwCampbell what kind of improvements could I do to this PR such that it can be merged?

HuwCampbell:
Look, I'm not sure right now. While it's a minor inconvenience to add `mconcat`, I do think that it's probably a better API to accept a list.

But I can't make breaking changes like that in the next release. The last release was forced on me due to like, one of our two dependencies deprecating without due diligence, but it was just awful... absolutely harrowing.

PR #492, opened 2024-08-31
Add maybeValue combinator.

alunduil:
This combinator is a convenience for handling default values that may
fail. I use this pattern to determine sane defaults from the system if
available but if not enforce the value to be required instead of having
a default.

I'm not sure if this is the best location nor what the recommended
testing strategy is for something like this, but please let me know if
there is interest in adding a new combinator such as this and what I can
improve to ensure it's successfully merged.


HuwCampbell:
G'day,

Thanks for thinking about optparse.

While this might be useful, I don't feel like it's quite common enough that people will reach for it over just using the constituents. Once inlined it's just

```haskell
maybe mempty value mx <> showDefault
```

or even

```haskell
foldMap value mx <> showDefault
```

PR #494, opened 2024-09-19
Allow to tune behavior of the brief description renderer

smelc:
# Context

First of all, thank you @HuwCampbell for your work on `optparse-applicative`. We (at [cardano-cli](https://github.com/input-output-hk/cardano-cli)) are very happy users of `optparse-applicative`!

This PR allows to tune the behavior of the renderer of descriptions. An option is added that favors vertical alignment when displaying multiple flags, instead of maximizing the use of space on a single line.

The default behavior is unchanged and no change is imposed upon users. This PR solely make it possible for callers that prefer vertical alignment to do so.

# Caveats

I am not super savvy about the different options that could be proposed. I have only proposed the existing behavior and the new one, which happen to yield very different results, as the change to the golden file of the changed test exemplifies:

image

We have many more examples at `cardano-cli` where we find this change beneficial in terms of readibility of the `--help` message when there are many flags available. For example:

image

Here is another one:

image

# History

For the record, this is an improvement that originally comes from https://github.com/pcapriotti/optparse-applicative/pull/428 and which we have maintained in [this fork](https://github.com/input-output-hk/optparse-applicative), but we now want to remove this fork, to shrink the spread in the Haskell ecosystem, and reduce maintenance burden.

### DIFF
--- a/src/Options/Applicative/Builder.hs
+++ b/src/Options/Applicative/Builder.hs
@@ -90,6 +90,7 @@ module Options.Applicative.Builder (
   helpLongEquals,
   helpShowGlobals,
   helpIndent,
+  briefHangPoint,
   prefs,
   defaultPrefs,
 
@@ -582,6 +583,9 @@ helpShowGlobals = PrefsMod $ \p -> p { prefHelpShowGlobal = True }
 helpIndent :: Int -> PrefsMod
 helpIndent w = PrefsMod $ \p -> p { prefTabulateFill = w }
 
+-- | Set the width at which to hang the brief help text.
+briefHangPoint :: Int -> PrefsMod
+briefHangPoint php = PrefsMod $ \p -> p { prefBriefHangPoint = php }
 
 
 -- | Create a `ParserPrefs` given a modifier
@@ -597,7 +601,8 @@ prefs m = applyPrefsMod m base
       , prefColumns = 80
       , prefHelpLongEquals = False
       , prefHelpShowGlobal = False
-      , prefTabulateFill = 24 }
+      , prefTabulateFill = 24
+      , prefBriefHangPoint = 35 }
 
 -- Convenience shortcuts
 
--- a/src/Options/Applicative.hs
+++ b/src/Options/Applicative.hs
@@ -202,6 +202,7 @@ module Options.Applicative (
   helpLongEquals,
   helpShowGlobals,
   helpIndent,
+  briefHangPoint,
   defaultPrefs,
 
   -- * Completions
--- a/src/Options/Applicative/Help/Core.hs
+++ b/src/Options/Applicative/Help/Core.hs
@@ -414,7 +414,7 @@ parserUsage pprefs p progn =
     hsep
       [ pretty "Usage:",
         pretty progn,
-        hangAtIfOver 9 35 (extractChunk (briefDesc pprefs p))
+        hangAtIfOver 9 (prefBriefHangPoint pprefs) (extractChunk (briefDesc pprefs p))
       ]
 
 -- | Peek at the structure of the rendered tree within.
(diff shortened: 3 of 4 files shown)


HuwCampbell:
G'day,

So if memory servers `hangAtIfOver` becomes exactly `align` if the column is less than the 35 parameter, so it might be even simpler to tune that what you have.

I'll have a proper look and think soon, I'm not too opposed to this if folks would find it useful, though to be honest I think the default layout is nicer.

Cheers.

smelc:
Nice to meet you @HuwCampbell and looking forward to your review :sunglasses:

smelc:
Hey @HuwCampbell, I hope you're doing good :wave: Did you have the time to look at this PR?

HuwCampbell:
G'day.

Yes, I am doing quite well. My family has a new puppy, they're very time consuming (both my kids and the puppy).

Honestly I'm a bit finicky about changes to the pretty printing because there's no one size that fits all. I just try to balance things as best I can to make our users' users experience as pleasant as possible.

Personally, I think that pushing everything over to the right that far and having alternatives split over different lines like you've shown is quite undesirable and should be avoided. The printer takes quite a bit of care to keep alternatives on the same line if it can. (This is the functions `groupOrNestLine` as part of `foldTree` over alternatives, I think it does this really well).

That said, I'm not going to unequivocally say no to a simple customisation if it's called for.

As far as this PR goes, I don't think this is quite the right thing. If I ever want to adjust said option it's a PVP breaking change right on the data type, and I try to do at most one a year at this stage.

So I would prefer just one integer option on the level at which you prefer to reïndent.

src/Options/Applicative/Builder.hs
```diff
@@ -582,6 +583,9 @@ helpShowGlobals = PrefsMod $ \p -> p { prefHelpShowGlobal = True }
 helpIndent :: Int -> PrefsMod
 helpIndent w = PrefsMod $ \p -> p { prefTabulateFill = w }
 
+-- | Set the pretty printer for brief help text.
+briefPrettyPrinter :: BriefPrettyPrinter -> PrefsMod
+briefPrettyPrinter bpp = PrefsMod $ \p -> p { prefBriefPP = bpp }
```
HuwCampbell, line 588: I think it would be better to do this differently, requiring the consumer of optparse to use the constructor and having it exported in types makes it a part of the "public API" which is a bit fragile, and not really in line with the other mods we have.

In my opinion it's better to instead have an option like:
```haskell
briefHangPoint :: Int -> PrefsMod
```

which sets the location where the hang constructor kicks in.


Now, when I mentioned that I thought the constructor was redundant, I was going off the definition of `hangAtIfOver` being

```haskell
hangAtIfOver :: Int -> Int -> Doc -> Doc
hangAtIfOver i j d =
  column $ \k ->
    if k <= j then
      align d
    else
      linebreak <> ifAtRoot (indent i) d
```
so if one sets this to maxBound it is functionally identical to `align`.

smelc: Hey @HuwCampbell, I did the change and indeed it's simpler :+1:

And I confirm that setting `j` to `maxBound` is indeed functionally equivalent to the previous design :+1:

tests/formatting-long-subcommand.err.txt
```diff
@@ -1,7 +1,8 @@
-Usage: formatting-long-subcommand hello-very-long-sub 
-         [TARGET...] [--first-flag] 
-         [--second-flag] [--third-flag] 
-         [--fourth-flag]
+Usage: formatting-long-subcommand hello-very-long-sub [TARGET...]
+                                                      [--first-flag]
+                                                      [--second-flag]
+                                                      [--third-flag]
+                                                      [--fourth-flag]
```
HuwCampbell, line 5: The point of this test was to show the behaviour of the new hang feature, so if you could delete the second commit I'd be much obliged.

smelc: @HuwCampbell> I deleted the second commit and rebased on latest `master` :+1:

Locally on my machine, all tests pass.

smelc:
@HuwCampbell> do you have a clue of when you will do a release of `optparse-applicative`? (asking to have an indication of when we will be able to use this PR's feature without maintaining our fork).

Hope you're doing great!

HuwCampbell:
Release candidate is up here:
https://hackage.haskell.org/package/optparse-applicative-0.19.0.0/candidate

PR #498, opened 2025-03-08
Remove transformers-compat dependency

josefs:
It's not used anymore


HuwCampbell:
Yeah, I think it's time too. But we should also bump transformers to something within the 2020s too.

HuwCampbell:
Ta.

PR #500, opened 2025-04-03
README: Use `hsubparser` in example code

9999years:
Example code should use `hsubparser` instead of `subparser` because generating good `--help` text is one of the main responsibilities of a CLI argument parsing framework.

I was using the example code and found myself very confused that `foo cmd --help` would ignore `cmd` entirely but `foo --help cmd` would work correctly. This behavior is extremely unintuitive and should not be the default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,7 +555,7 @@ on each command), and commands can be added with the `command`
 modifier. For example,
 
 ```haskell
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Ta.

Issue #491, opened 2024-08-16
Support for ospath/osstring


amigalemming:
Can we have support for `OsPath`/`OsString`?

The idea of those types is to avoid converting from system (filepath) strings to `String` and back. But currently this is what happens if we want to construct an argument parser that returns `OsPath`.

tbidne:
I, too, would like this. It's easy enough to provide the conversion on top of the string reader:

```haskell
osPath :: ReadM OsPath
osPath = OA.str >>= \fp -> case OsPath.encodeUtf fp of
  Right txt -> pure txt
  Left ex -> fail $ "failed: " ++ displayException ex
```

But I imagine you want to skip the conversion entirely. Seems possible to me, but there are some tricky decisions.

## 1. Implement new getArgs'

This would require replacing `optparse`'s usage of `getArgs :: IO [String]` with `getArgs' :: IO [OsString]`. Fortunately, the function itself is pretty straightforward. The `unix` library already does this for `PosixString`, which differs from `base`'s `getArgs` only in the final conversion:

```haskell
getArgs :: IO [PosixString]
getArgs =
  alloca $ \ p_argc ->
  alloca $ \ p_argv -> do
   getProgArgv p_argc p_argv
   p    <- fromIntegral <$> peek p_argc
   argv <- peek p_argv
   -- base uses 'peekCString :: TextEncoding -> CString -> IO String'
   peekArray (p - 1) (advancePtr argv 1) >>= mapM (fmap PS . B.packCString)
```

So, assuming we can just do the same thing for `WindowsString`, we'd have:

```haskell
getArgs' :: IO [OsString]
getArgs' =
  alloca $ \ p_argc ->
  alloca $ \ p_argv -> do
   getProgArgv p_argc p_argv
   p    <- fromIntegral <$> peek p_argc
   argv <- peek p_argv
#if WINDOWS
   peekArray (p - 1) (advancePtr argv 1) >>= mapM (fmap (OsString . WS) . B.packCString)
#else
   peekArray (p - 1) (advancePtr argv 1) >>= mapM (fmap (OsString . PS) . B.packCString)
#endif
```

Armed with this, you have two choices:

1. Replace `ReadM = ReaderT String (Except ParseError)` with `ReadM = ReaderT OsString (Except ParseError)`.

    This is a **breaking** change because `ReadM` is fully exposed. It's unfortunate, but most of `optparse`'s API doesn't expose these internals, so maybe it's palatable.
    
[...]

PR #504, opened 2025-07-01
Initial port to os-string

0rphee:
- Tests pass (Though they have changed since most of the API is not longer in terms of `String`)
- Adjusted build-depends bounds and remove support to ghc <8.6 (due to `process >= 1.6.26.0`). Also, as mentioned in #491, `filepath >= 1.5` and `os-string >= 2.0.5` are needed.
  - This dependencies prevent the use of this new version in packages which depend on `ghc` as mentioned in #491.
- Since `text` is already a transitive dependency, so I also updated some of the API to use that instead of `String`.
  - As the output of `prettyprinter` is `LazyText` I've updated the API to not change this type.
  - I use `Text.IO` functions for printing.
- I'm not entirely sure of some of the encoding/decoding stuff, I've used `unsafePerformIO . decodeLE`, but also new functions `Options.Applicative.Help.Pretty.osStringToStrictText` (built in terms of `ByteString -> Text` functions). It would be nice if there was a straight `OsString -> Text`, instead of the need to pass through `String` or `ByteString`.
- For the moment I've left `Options.Applicative.Help.Levenshtein` as-is (using `String`), but the `Options.Applicative.Builder.Completer` (which has heavy use of `uncons`) module is updated, it's expected that `Levenshtein` would look like `Completer`.
- It'd be nice to have `OsString.getProgName` and `OsString.readProcess` in `process`.


0rphee:
@HuwCampbell hi, sorry for the ping. Is OsString support planned? or should it be better to make a different package?

hasufell:
@0rphee if upstream doesn't respond, then I'd be willing to maintain a fork.

0rphee:
@hasufell sorry for the late response. 

At the moment, and for the foreseeable future, I don't think that I am able to help in such endeavor. However, I fully support your intention to create a new package. 

Regarding the code, I tried my best to keep everything as-is, but some of my biggest concerns are:

1. The distance algorithms since, if I remember correctly, I translate `OsString -> String`, and use the original code mostly unchanged. 

2. The API. At first I thought it was better to keep most things as a drop in replacement, however, there were some changes that don't keep the change String-OsString 1:1. Maybe it would be nice to separate the OsString interface to a more primitive API, and on top of that, implement the previous String-based API. However, I imagine this is considerably more work than just maintenance.

Issue #352, opened 2019-07-05
Selective Applicative Functor Instance?


HuwCampbell:
Just a question at this stage.

What would adding a `Selective` instance do to the user interface? I actually hypothesise that we already sort of have one, and that the current commands interface is a selective applicative. So the current commands implementation could be rewritten to also support things like this:

```
Available commands:
  hello                    Selected with a positional token
  --this                   Selected with a flag
```

HuwCampbell:
I have implemented this and it's underwhelming.

Given:
```haskell
data Value = B | C String
  deriving (Eq, Show)

a :: Parser String
a = strOption (short 'a')

b :: Parser Value
b = flag' B (short 'b')

c :: Parser (String -> Value)
c = flag' C (short 'c')

d :: Parser Value
d = ((Left <$> a) <|> (Right <$> b)) <*? c
```

We can expect this:
```bash
# Just to be clear, how this is rendered is just my current implemenation. It's hard to express clearly.
:main --help
Usage: <interactive> (-a ARG | -b) [-c]

:main -a ok
Missing: -c

:main -a ok -c
C "ok"

:main -b
B

:main -c
Missing: (-a ARG | -b)

:main -c -b
B
```

The first 3 cases make sense to me. The last two are problematic. Providing a `-c` yields a message saying I need an `a` or a `b`, but I won't be used by the `b` at all anyway. `-b -c` is exactly the same as `-b`.


I think most would use
```haskell
d :: Parser Value
d = a <**> c <|> b
```

merijn:
Is the code of this selective attempt still somewhere? I think there's some nice use-cases for it and if I don't have to figure the implementation out myself, that saves a lot of time :)

HuwCampbell:
https://github.com/pcapriotti/optparse-applicative/tree/expoloration/selective

googleson78:
I've also been thinking about this, and it feels like having a selective interface makes perfect sense for subcommands. Did your original exploration try something like this?

PR #505, opened 2025-07-05
revive the idea of supporting Selective

exaexa:
Hello,

this PR is essentially the branch mentioned by @HuwCampbell in issue #352 with minor updates (rebased, versions bumped, and some minor cleaning).

I think it would be great to have Selectives in optparse-applicative, in order to support various sensible ways of defaulting the options -- my target use-case is specifying input/output file formats, which may be guessed from the filename suffix (in which case the format option isn't required), but should throw an error if the guess is impossible. The selective interface allows handling of such error very nicely within the optparse error-printing scheme, instead of having to print a custom error message after parsing.

Notably, similar functionality is possible with the `ParserM` interface, but it's somewhat ugly for usual Monad vs. Selective purposes (the "continuation" branch requires a parameter to materialize, so it can't really get analyzed statically, thus there's no general way to get help for the options out of that).

An example for why this is IMO good is:

```hs
import Control.Selective
import Data.List
import Options.Applicative

guess :: String -> Maybe String -> Either String (String, String)
guess x ty'
  | Just ty <- ty' = Right (ty, x)
  | ".md" `isSuffixOf` x = Right ("markdown", x)
  | ".png" `isSuffixOf` x = Right ("picture", x)
  | otherwise = Left x

opts =
  select
    (guess
       <$> strOption (long "filename" <> help "input")
       <*> optional
             (strOption
                $ long "type"
                    <> metavar "TYPE"
                    <> hidden
                    <> help "input type (defaulted from file name)"))
    ((,) <$> strOption (long "type" <> metavar "markdown|picture"))

main = execParser (info (opts <**> helper) mempty) >>= print
```

Help prints out quite sensibly as:
```
ghci> :main -h
Usage: <interactive> --filename ARG [--type markdown|picture]

Available options:
  --filename ARG           input
  --type TYPE              input type (defaulted from file name)
```

In the example, the argument is defaulted just right:
```
ghci> :main --filename test.png
("picture","test.png")
```

In case of unknown filetype the error message is relatively sensible:
```
ghci> :main --filename test.jpg
Missing: --type markdown|picture

Usage: <interactive> --filename ARG [--type markdown|picture]
```

```
ghci> :main --filename test.jpg --type picture
("picture","test.jpg")
```

One can ofc juggle the options a bit (e.g., hide various subsets of them in various situations) to achieve variations on the final effect.

On a minor note, I'm not sure if the implementation is completely OK -- I found that e.g. this https://github.com/exaexa/optparse-applicative/commit/cb2344f3de3579a199830014eaf85a730f60647c#diff-0881f202d8624335f28737091abea8a17f3e203e5e8e3ca398764fde24e91b93R153 works as well and I wasn't able to find a testcase that would help me understand if one of these would be strictly better. Any advice on that very welcome.

I can write a little bit of documentation; e.g. some simplified version of the above example could go to the Selective instance. If desirable, please let me know, will do.

Thanks!
-mk

### DIFF
--- a/optparse-applicative.cabal
+++ b/optparse-applicative.cabal
@@ -114,6 +114,7 @@ library
                      , transformers                    >= 0.5   && < 0.7
                      , prettyprinter                   >= 1.7   && < 1.8
                      , prettyprinter-ansi-terminal     >= 1.1.2 && < 1.2
+                     , selective                       >= 0.4   && < 0.8
 
   if flag(process)
     build-depends:     process                         >= 1.0   && < 1.7
--- a/src/Options/Applicative/Common.hs
+++ b/src/Options/Applicative/Common.hs
@@ -55,6 +55,7 @@ import Control.Applicative
 import Control.Monad (guard, mzero, msum, when)
 import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.State (StateT(..), get, put, runStateT)
+import Control.Selective (select)
 import Data.List (isPrefixOf)
 import Data.Maybe (maybeToList, isJust, isNothing)
 import Prelude
@@ -270,7 +271,7 @@ evalParser (NilP r) = r
 evalParser (OptP _) = Nothing
 evalParser (MultP p1 p2) = evalParser p1 <*> evalParser p2
 evalParser (AltP p1 p2) = evalParser p1 <|> evalParser p2
-evalParser (SelectP p k) = evalParser p >>= either (\a -> ($ a) <$> evalParser k) pure
+evalParser (SelectP p k) = select (evalParser p) (evalParser k)
 evalParser (BindP p k) = evalParser p >>= evalParser . k
 
 -- | Map a polymorphic function over all the options of a parser, and collect
--- a/src/Options/Applicative/Common.hs
+++ b/src/Options/Applicative/Common.hs
@@ -147,6 +147,16 @@ searchParser f (MultP p1 p2) = foldr1 (<!>)
 searchParser f (AltP p1 p2) = msum
   [ searchParser f p1
   , searchParser f p2 ]
+searchParser f (SelectP p k) = msum
+  [ do p' <- searchParser f p
+       return $ SelectP p' k
+  , case evalParser p of
+      Nothing -> do
+        k' <- searchParser f k
+        return $ SelectP p k'
+      Just (Left a) -> fmap ($ a) <$> searchParser f k
+      Just (Right _) -> mzero
+  ]
 searchParser f (BindP p k) = msum
   [ do p' <- searchParser f p
        return $ BindP p' k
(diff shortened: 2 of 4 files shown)


HuwCampbell:
Hi,
Let's have a look at that diff.

```haskell
searchParser f (SelectP p k) = msum
  [ do p' <- searchParser f p
       return $ SelectP p' k
  , do k' <- searchParser f k
       return $ SelectP p k'
  ]
```

```haskell
searchParser f (SelectP p k) = msum
  [ do p' <- searchParser f p
       return $ SelectP p' k
  , case evalParser p of
      Nothing -> do
        k' <- searchParser f k
        return $ SelectP p k'
      Just (Left a) -> fmap ($ a) <$> searchParser f k
      Just (Right _) -> mzero
  ]
```

The difference here is that as we step through the command line, do we want to
stop people entering the second option if the first is chosen and reifies as a `Right`.

This version supports permutation and the potential for parsing an item which won't
ever be used, but there's actually a third option as well, which is this:
```haskell
searchParser f (SelectP p k) = msum
  [ do p' <- searchParser f p
       return $ SelectP p' k
  , case evalParser p of
      Just (Left a) -> fmap ($ a) <$> searchParser f k
      _ -> mzero
  ]
```
which forces the left to be completed before the second part can be considered.

Those two extremes are more like `selectA` and more like `selectM` from the paper as far
as I can tell.

As an aside, your initial example permits this:
```
ghci> :main --type picture --type markdown --filename as.md 
("picture","as.md")
```
with the implementation in the middle, but not with the last. The first (more applicative
version) supports this as well

```
ghci> :main --type picture --filename as.md --type markdown 
("picture","as.md")
```

so maybe the `selectM` like option is the best? I don't really have a feel for it at this
stage. With the `selectM` style we can't permit permutation; on the other side we
allow items which are completely ignored.


With regards to your example, it's annoying that you have to resort to hacky things
like `hidden` modifiers.

Nevertheless, if I add a Monad instance, this is identical both functionally
and for the help text to what you have.

```haskell
opts =
  selectM
    (guess
       <$> strOption (long "filename" <> help "input")
       <*> optional
             (strOption
                $ long "type"
                    <> metavar "markdown|picture"
                    <> help "input type (defaulted from file name)"))
    ((,) <$> strOption (long "type" <> metavar "markdown|picture"))
```
where I've removed the `hidden` modifier.

It would be good to search for some examples that are more idiomatic maybe?

exaexa:
> It would be good to search for some examples that are more idiomatic maybe?

OK, will try. I thought having the full selective "if" there (was it `branch`?) could be more powerful but need to check if that actually does help anything.

With the `hidden` -- it would be great if the option renderer was able to correctly render the "conditional" commands as conditional with some given condition. Absolutely unsure of how the API (or the rendering!) for that would look like; I assume extra examples for this will help to map the design space.

Anyway thanks for the points about the repeated options! These are indeed an issue that I overlooked.

( From the higher perspective, it seems to me that the use I'm looking for is very close to the commands and subcommands (these technically feel "selective" -- if a subcommand is parsed out, then suddenly more options appear), except the "command" parser is not a command but some value-dependent option combination. )

exaexa:
OK so I've had this on a 10day snooze to see if I find a clearer usecase, but it seems that there's no simple one at this point. I'll see how this ages and come back if something surfaces; in the meantime I guess we can close this here to prevent crufting the PR list. Thanks again for the elaboration here!

PR #512, opened 2025-12-16
Make it compile with MicroHs

augustss:
The title say it all

### DIFF
--- a/src/Options/Applicative/BashCompletion.hs
+++ b/src/Options/Applicative/BashCompletion.hs
@@ -11,7 +11,7 @@ module Options.Applicative.BashCompletion
     zshCompletionScript,
   ) where
 
-import Control.Applicative
+import Control.Applicative hiding(asum)
 import Prelude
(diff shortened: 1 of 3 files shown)


src/Options/Applicative/BashCompletion.hs
```diff
@@ -13,7 +13,6 @@ module Options.Applicative.BashCompletion
 
 import Control.Applicative
 import Prelude
-import Data.Foldable ( asum )
```
HuwCampbell, line 16: This is required for GHC 9.0 and below, which we still support.

HuwCampbell:
Yeah rad, I didn't realise it was that close already. Last time I tried I got a bit stuck on `text` and the pretty printer.

If it's going to be supported we should consider CI as well.

augustss:
I've changed the PR so it works with old GHC and mhs.

I'll add CI as well.

HuwCampbell:
Ta

PR #514, opened 2026-01-29
remove fullDesc from README example

exaexa:
Changelog for 0.20 says that `fullDesc` is removed (and didn't really do much for a few releases). This simply removes it from the README example, so that people who install 0.20 now for any reason (like me, naively pulling stuff from git :D ) don't get surprised.

Thanks!


HuwCampbell:
Ta

Issue #484, opened 2023-07-28
Default command and command alias 


bingis-khan:
Is it possible to create an alias of a command? For example in the help menu, under "Available commands", there would be something like this:
```
Available commands:
  search, s                         Do some searching or whatever.
  ...
```

Also might be related: is it possible to specify a defaut command?

---

Note that all of the things above are possible with a liberal use of Alternative, but these make the help menu much less appealing:
```
                                    \/-- looks meh
Usage: surcher [(-p|--port PORT) | COMMAND] [-d|--dir PATH]

Available options:
  -p,--port PORT           Port on which to serve the web interface.  <- this should probably be only visible in the help menu for "serve"
  -h,--help                Show this help text

Available commands:
  search                   Execute a single search with a query.
  s                        Serve a website for searching. <- """""alias"""""
  serve                    Serve a website for searching.
  ```

**Comment by HuwCampbell** [comment]
Mmm.

So I get that you want this:

```
Usage: surcher [(-p|--port PORT) | COMMAND] [-d|--dir PATH]

Available options:
  -p,--port PORT           Port on which to serve the web interface.  <- this should probably be only visible in the help menu for "serve"
  -h,--help                Show this help text

Available commands:
  search                   Execute a single search with a query.
  serve, s                 Serve a website for searching.
```

But what were you thinking for how to represent the default? What should that look like?

Issue #367, opened 2019-11-24
Create a harder split between internal and public modules


HuwCampbell:
We're currently doing an explicit export list from `Options.Applicative` and `Options.Applicative.Arrows` which is really the complete public API.

Perhaps everything else should be scoped into an `Internal` hierarchy with haddock not home annotations on them.

PR #515, opened 2026-04-14
Implement command aliases

tbidne:
This implements the "command aliases" portion of #484 (the other part, default commands, is left for another day). My motivation is presumably similar to the OP of that issue i.e. I sometimes want shorter aliases for commands, much like we can do with options. For example:

```haskell
-- options
long "file"
  <> short 'f'

-- new, for commands
-- commandAliases :: NonEmpty String -> ParserInfo a -> Mod CommandFields a
commandAliases
  ("print" :| ["p"])
  (info (pure Print) (progDesc "Runs print"))
```

I chose to implement this as a companion to the normal function:

```haskell
command :: String -> ParserInfo a -> Mod CommandFields a
```

since `CommandFields` is a modifier for the command group, not an individual command, so we cannot monoidally combine names for an individual command, like we can for options.

This is technically a breaking change as `Options.Applicative.Builder.Internal` exports the `CommandFields` constructor:

```diff
diff --git a/src/Options/Applicative/Builder/Internal.hs b/src/Options/Applicative/Builder/Internal.hs
index 2110067..68984c2 100644
--- a/src/Options/Applicative/Builder/Internal.hs
+++ b/src/Options/Applicative/Builder/Internal.hs
 data CommandFields a = CommandFields
-  { cmdCommands :: [(String, ParserInfo a)]
+  { cmdCommands :: [(NonEmpty String, ParserInfo a)]
   , cmdGroup :: Maybe String }
```

Incidentally, it would be great to have #367, as imo this should really only require a minor bump, not a major one. I would be happy to investigate that, though there is less need for it wrt this PR, since we are already on a new major bump.

Thanks!

### DIFF
--- a/src/Options/Applicative/BashCompletion.hs
+++ b/src/Options/Applicative/BashCompletion.hs
@@ -17,10 +17,11 @@ import Prelude
 #if !defined(__MHS__)
 import Data.Foldable ( asum )
 #endif
+
+import qualified Data.Foldable as Foldable
 import Data.List ( isPrefixOf )
 import Data.List.NonEmpty (NonEmpty)
-import qualified Data.List.NonEmpty as NE
-import Data.Maybe ( fromMaybe, listToMaybe )
+import Data.Maybe ( fromMaybe, listToMaybe, mapMaybe )
 
 import Options.Applicative.Builder
 import Options.Applicative.Common
@@ -159,12 +160,20 @@ bashCompletionQuery pinfo pprefs richness ws i _ = case runCompletion compl ppre
     --
     -- and the user types 'r', we should return
     --
-    --   [(retry, p1), (run, p2), (r, p2)]
+    --   [(retry, p1), (run, p2)]
+    --
+    -- If the first entry (command name) is the canonical one, we should
+    -- favour it in completions, as if we were to provide all options there
+    -- would be a good chance we'd be forcing the user to tab through and
+    -- disambiguate between functionally identical options.
+    --
+    -- In zsh and fish we also provide the help doc in the completions, which
+    -- we don't want to repeat a whole bunch of times.
     filter_completions :: [(NonEmpty String, ParserInfo a)] -> [(String, ParserInfo a)]
     filter_completions =
-      let matchZip :: (NonEmpty String, b) -> [(String, b)]
-          matchZip (aliases, painfo) = (\x -> (x, painfo)) <$> NE.filter is_completion aliases
-      in (>>= matchZip)
+      let findAlias :: (NonEmpty String, b) -> Maybe (String, b)
+          findAlias (aliases, painfo) = (\x -> (x, painfo)) <$> Foldable.find is_completion aliases
+      in mapMaybe findAlias
 
     -- We only want to show a single line in the completion results description.
     -- If there was a line break, it would come across as a different completion
--- a/src/Options/Applicative/BashCompletion.hs
+++ b/src/Options/Applicative/BashCompletion.hs
@@ -18,6 +18,8 @@ import Prelude
 import Data.Foldable ( asum )
 #endif
 import Data.List ( isPrefixOf )
+import Data.List.NonEmpty (NonEmpty)
+import qualified Data.List.NonEmpty as NE
 import Data.Maybe ( fromMaybe, listToMaybe )
 
 import Options.Applicative.Builder
@@ -120,7 +122,7 @@ bashCompletionQuery pinfo pprefs richness ws i _ = case runCompletion compl ppre
          | argumentIsUnreachable reachability
         -> return []
          | otherwise
-        -> return . with_cmd_help $ filter (is_completion . fst) ns
+        -> return . with_cmd_help $ filter_completions ns
 
     -- When doing enriched completions, add any help specified
     -- to the completion variables (tab separated).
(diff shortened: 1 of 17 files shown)


tbidne:
Some bikeshedding notes:

1. Currently, this differs from the proposed rendering in the issue i.e. we have *no* whitespace between aliases:

    ```
    Usage: prop_cmd_aliases (COMMAND | COMMAND)

    Available options:

    Available commands:
      hello,hi                 Print greeting
      goodbye                  Say goodbye

    French commands:
      bonjour                  Print greeting
      au-revoir,adieu,ciao     Say goodbye

    Other commands:
      health                   Check health
      aux                      Auxiliary
    ```
    
    This is for consistency with options (e.g. see `-h,--help` above), though it is arguably busy compared to the variant with a single space:

    ```
    Usage: prop_cmd_aliases (COMMAND | COMMAND)

    Available options:

    Available commands:
      hello, hi                Print greeting
      goodbye                  Say goodbye

    French commands:
      bonjour                  Print greeting
      au-revoir, adieu, ciao   Say goodbye

    Other commands:
      health                   Check health
      aux                      Auxiliary
    ```

2. There is no normalizing wrt repeated aliases, for instance, duplicate aliases will be listed in completions multiple times. This is consistent with how commands currently work e.g.

    ```haskell
    command "cmd" (info (pure Cmd1) ...)
      <> command "cmd" (info (pure Cmd2) ...)
    ```
    
    Both will be listed and CLI `cmd` will choose `Cmd2`.

tbidne:
The microhs error is baffling me:

```
mhs: uncaught exception: error: "src/Options/Applicative/Common.hs": line 200, col 48: Cannot satisfy constraint: ([] ~ NonEmpty)
     fully qualified: (Primitives.~ Data.List_Type.[] Data.List.NonEmpty_Type.NonEmpty)
```

It is complaining [here](https://github.com/pcapriotti/optparse-applicative/pull/515/changes#diff-0881f202d8624335f28737091abea8a17f3e203e5e8e3ca398764fde24e91b93R200):

```
    cmdMatches cs
      | prefDisambiguate prefs = snd <$> filter (any (isPrefixOf arg) . fst) cs
      --                                 HERE: vvv
      | otherwise = maybeToList (lookupCmd arg cs)
```

The types are:

```haskell
cs :: [(NonEmpty String, ParserInfo r)]
lookupCmd :: String -> [(NonEmpty String, a)] -> Maybe a
```
:shrug:

HuwCampbell:
Pretty sure that error is because micro haskell has a different type for `any`.

```haskell
any :: forall a . (a -> Bool) -> [a] -> Bool
```

As opposed to it being over `Foldable`.
```haskell
any :: Foldable t -> (a -> Bool) -> t a -> Bool
```

src/Options/Applicative/Extra.hs
```diff
@@ -320,7 +321,7 @@ parserFailure pprefs pinfo msg ctx0 = ParserFailure $ \progn ->
               CmdReader _ ns    | argumentIsUnreachable reachability
                                -> []
                                 | otherwise
-                               -> fst <$> ns
+                               -> join $ NE.toList . fst <$> ns
```
HuwCampbell: Use bind here please

src/Options/Applicative/Builder.hs
```diff
@@ -240,7 +242,28 @@ style x = optionMod $ \p ->
...
+-- @
+--
+-- > Available commands:
+-- >   hello, hi                Print greeting
+-- >   goodbye                  Say goodbye
+--
+-- @since 0.20.0.0
+commandAliases :: NonEmpty String -> ParserInfo a -> Mod CommandFields a
+commandAliases aliases pinfo = fieldMod $ \p ->
+  p { cmdCommands = (aliases, pinfo) : cmdCommands p }
```
HuwCampbell, line 266: Maybe `commandWithAliases`?

To my ear, `commandAliases` sounds like an accessor.

HuwCampbell (approved):
Excellent as usual. Thanks.

I'll do some tyre kicking, but I see no current issues.

src/Options/Applicative/Extra.hs
tbidne: Ah, good call.

src/Options/Applicative/Builder.hs
tbidne: Fine with me.

tbidne:
> Pretty sure that error is because micro haskell has a different type for `any`.

Bang on, fixed now.

Thanks for the review and kind words!

src/Options/Applicative/BashCompletion.hs
```diff
@@ -148,6 +150,22 @@ bashCompletionQuery pinfo pprefs richness ws i _ = case runCompletion compl ppre
...
+    --   - ([search], p3)
+    --
+    -- and the user types 'r', we should return
+    --
+    --   [(retry, p1), (run, p2), (r, p2)]
+    filter_completions :: [(NonEmpty String, ParserInfo a)] -> [(String, ParserInfo a)]
+    filter_completions =
+      let matchZip :: (NonEmpty String, b) -> [(String, b)]
+          matchZip (aliases, painfo) = (\x -> (x, painfo)) <$> NE.filter is_completion aliases
+      in (>>= matchZip)
```
HuwCampbell: Just spit balling here, but maybe

```haskell
    filter_completions :: [(NonEmpty String, ParserInfo a)] -> [(String, ParserInfo a)]
    filter_completions =
      let matchZip :: (NonEmpty String, b) -> Maybe (String, b)
          matchZip (aliases, painfo) = (\x -> (x, painfo)) <$> find is_completion aliases
      in mapMaybe matchZip
```

would be more appropriate.

HuwCampbell: Yeah, I'm pretty sure just returning the first option is the right thing to do here.

- It's more likely that aliases will share a prefix, so we'd be demanding the user do more work (selecting between functionally identical items).
- The first one is probably preferred
- In zsh and fish we don't want to repeat the command help text a bunch of times.

HuwCampbell: I've pushed something along those lines; let me know what you think.

tbidne: Apologies for the delay. That makes sense, thanks for the change.

tbidne:
Thanks for the merge!

